PR #2, opened 2024-06-21
initial charm

dstathis:
Brand new charm!


charmcraft.yaml
ca-scribner, line 1: I'm all for using the combined metadata/charmcraft.yaml file, but atm I see a [few places using metadata.yaml](https://github.com/search?q=repo%3Acanonical%2Fobservability%20metadata&type=code) in the reused workflow code so I think we'd need to fix those as well.

ca-scribner: afaict they only matter for release and lib bumps, which wont be so important for a little while here.  So we can do those fixes over time and it doesn't have to block us from merging this.

if you're +1 on using the combined file, I'll do the same in istio-core

pyproject.toml
```diff
@@ -0,0 +1,48 @@
...
+[tool.coverage.run]
+branch = true
+
+[tool.coverage.report]
+show_missing = true
+
+[tool.pytest.ini_options]
+minversion = "6.0"
+log_cli_level = "INFO"
+
```
ca-scribner, line 15: I think this solves a "in the future, this is going to break" warning in CI.  But I don't recall the details and might be wrong about it

```suggestion
log_cli_level = "INFO"
asyncio_mode = "auto"

```

tox.ini
```diff
@@ -0,0 +1,93 @@
...
+set_env =
+    PYTHONPATH = {tox_root}/lib:{[vars]src_path}
+    PYTHONBREAKPOINT=pdb.set_trace
+    PY_COLORS=1
+pass_env =
+    PYTHONPATH
+    CHARM_BUILD_DIR
+    MODEL_SETTINGS
+
+[testenv:format]
```
ca-scribner, line 26: I think most obs repos use `fmt` here, but the chamcraft init uses format.  I'm not against either, but raising in case it wasn't an intentional difference.
```suggestion
[testenv:fmt]
```

tox.ini
```diff
@@ -0,0 +1,93 @@
...
+description = Apply coding style standards to code
+deps =
+    black
+    ruff
+commands =
+    black {[vars]all_path}
+    ruff --fix {[vars]all_path}
```
ca-scribner, line 33: ruff is going to deprecation warning you to death :)

```suggestion
    ruff check --fix {[vars]all_path}
```

README.md
ca-scribner, line 1: Should this repo be `istio-beacon` or `istio-beacon-operator`?  I see operator in most of the names, but can't remember if that's the historical but we don't want to anymore way, or the go forward way

tox.ini
```diff
@@ -0,0 +1,93 @@
...
+    coverage report
+
+[testenv:static-{charm,lib}]
+description = Run static type checks
+deps =
+    pyright
+    -r {tox_root}/requirements.txt
+commands =
+    charm: pyright {posargs} {[vars]src_path}
+    lib:
```
ca-scribner, line 78: I get why `lib: ` is left empty since there's no libs here, but it is so easy to forget to fill this out once a lib exists.  Do you know of any way to have this pass when there's no libs, but scan when there are libs?

README.md
dstathis: You are right. In fact it should be istio-beacon-k8s-operator. Fixed.

charmcraft.yaml
dstathis: Yeah we should use the combined file. It's the new way and we can fix CI to use it.

tox.ini
dstathis: `if [ -f "lib/istio-beacon-k8s/v*/*.py ]; then pyright ...; fi` would work but I feel this would be way uglier than I would like. Since I will be adding a lib very soon, I would prefer to just remember.

ca-scribner: Yeah agreed, leave it as is

ca-scribner (approved):

tox.ini
dstathis: Changed

PR #3, opened 2024-07-12
Mesh lib

dstathis:
## Issue
Create the service_mesh library

In scope for this PR: Using the constructor and relations to generate the relation data required to allow a service mesh to generate access policies. Adding apps to the mesh that relate over `service_mesh`.

In scope for the next PR: Add the cross_model_mesh interface support to allow for cross model relations.

Unrelated to the lib: Everything else. (model join for example has nothing to do with relations).

### DIFF
--- a/lib/charms/istio_beacon_k8s/v0/service_mesh.py
+++ b/lib/charms/istio_beacon_k8s/v0/service_mesh.py
@@ -1,0 +1,119 @@
+# Copyright 2024 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""#Service Mesh Library.
+
+The service mesh library is used to facilitate adding you charmed application to a service mesh.
+The library leverages the `service_mesh` interface and the `cross_model_mesh` interface
+
+To add service mesh support to your charm, instantiate a ServiceMeshConsumer object in the
+`__init__` method of your charm:
+
+```
+from charms.istio_beacon_k8s.v0.service_mesh import IsCrossModelError, Policy, ServiceMeshConsumer
+
+...
+self._mesh = ServiceMeshConsumer(
+    self,
+    policies=[
+        Policy(relation="logging", endpoints=[f"*:{HTTP_LISTEN_PORT}"]),
+    ],
+)
+```
+"""
+
+import json
+import logging
+import re
+from typing import List, Optional
+
+import pydantic
+from ops import CharmBase, Object
+
+LIBID = "3f40cb7e3569454a92ac2541c5ca0a0c"  # Never change this
+LIBAPI = 0
+LIBPATCH = 1
+
+logger = logging.getLogger(__name__)
+
+
+class Policy(pydantic.BaseModel):
+    """Data type for holding a service mesh policy."""
+
+    relation: str
+    endpoints: List[str]
+
+
+class ServiceMeshConsumer(Object):
+    """Class for managing the consumer side of the service_mesh relation interface."""
+
+    def __init__(
+        self,
+        charm: CharmBase,
+        mesh_relation_name: str = "service-mesh",
+        policies: Optional[List[Policy]] = None,
+    ):
+        """Class used for joining a service mesh.
+
+        Args:
+            charm: The charm instantiating this object.
+            mesh_relation_name: The relation name as defined in metadata.yaml or charmcraft.yaml
+                for the relation which used the service_mesh interface.
+            policies: List of access policies this charm supports.
+        """
+        super().__init__(charm, mesh_relation_name)
+        self._charm = charm
+        self._relations = self._charm.model.relations[mesh_relation_name]
+        self._policies = policies or []
+        self.framework.observe(
+            self._charm.on[mesh_relation_name].relation_created, self._relations_changed
+        )
+        self.framework.observe(self._charm.on.upgrade_charm, self._relations_changed)
+        for policy in self._policies:
+            self.framework.observe(
+                self._charm.on[policy.relation].relation_created, self._relations_changed
+            )
+            self.framework.observe(
+                self._charm.on[policy.relation].relation_broken, self._relations_changed
+            )
+
+    def _relations_changed(self, _event):
+        self.update_service_mesh()
+
+    def update_service_mesh(self):
+        """Update the service mesh.
+
+        Gathers information from all relations of the charm and updates the mesh appropriately to
+        allow communication.
+        """
+        logger.debug("Updating service mesh policies.")
+        policies = []
+        cmr_matcher = re.compile(r"remote\-[a-f0-9]+")
+        for policy in self._policies:
+            for relation in self._charm.model.relations[policy.relation]:
+                if cmr_matcher.fullmatch(relation.app.name):
+                    logger.debug(
+                        f"Cross model relation found: {relation.name}. Currently not implemented. Skipping."
+                    )
+                else:
+                    logger.debug(f"Found relation: {relation.name}. Creating policy.")
+                    policies.append(
+                        {
+                            "app_name": relation.app.name,
+                            "namespace": self._my_namespace(),
+                            "endpoints": policy.endpoints,
+                        }
+                    )
+        mesh_rel_data = {
+            "app_name": self._charm.app.name,
+            "model": self._my_namespace(),
+            "policies": policies,
+        }
+        for rel in self._relations:
+            rel.data[self._charm.app]["mesh_data"] = json.dumps(mesh_rel_data)
+
+    def _my_namespace(self):
+        """Return the namespace of the running charm."""
+        # This method currently assumes the namespace is the same as the model name. We
+        # should consider if there is a better way to do this.
+        return self._charm.model.name
--- a/lib/charms/istio_beacon_k8s/v0/service_mesh.py
+++ b/lib/charms/istio_beacon_k8s/v0/service_mesh.py
@@ -57,14 +57,20 @@ class Method(str, enum.Enum):
     trace = "TRACE"
 
 
-class Policy(pydantic.BaseModel):
-    """Data type for holding a service mesh policy."""
+class Endpoint(pydantic.BaseModel):
+    """Data type for a policy endpoint."""
 
-    relation: str
     hosts: List[str]
     ports: List[int]
     methods: List[Method]
     paths: List[str]
+
+
+class Policy(pydantic.BaseModel):
+    """Data type for holding a service mesh policy."""
+
+    relation: str
+    endpoints: List[Endpoint]
 
 
 class ServiceMeshConsumer(Object):
--- a/lib/charms/istio_beacon_k8s/v0/service_mesh.py
+++ b/lib/charms/istio_beacon_k8s/v0/service_mesh.py
@@ -3,7 +3,7 @@
 
 """#Service Mesh Library.
 
-The service mesh library is used to facilitate adding you charmed application to a service mesh.
+The service mesh library is used to facilitate adding your charmed application to a service mesh.
 The library leverages the `service_mesh` interface and the `cross_model_mesh` interface
 
 To add service mesh support to your charm, instantiate a ServiceMeshConsumer object in the
--- a/lib/charms/istio_beacon_k8s/v0/service_mesh.py
+++ b/lib/charms/istio_beacon_k8s/v0/service_mesh.py
@@ -92,7 +92,7 @@ class ServiceMeshConsumer(Object):
         """
         super().__init__(charm, mesh_relation_name)
         self._charm = charm
-        self._relation = self._charm.model.get_relation[mesh_relation_name]
+        self._relation = self._charm.model.get_relation(mesh_relation_name)
         self._policies = policies or []
         self.framework.observe(
             self._charm.on[mesh_relation_name].relation_created, self._relations_changed
@@ -115,6 +115,8 @@ class ServiceMeshConsumer(Object):
         Gathers information from all relations of the charm and updates the mesh appropriately to
         allow communication.
         """
+        if self._relation is None:
+            return
         logger.debug("Updating service mesh policies.")
         policies = []
         cmr_matcher = re.compile(r"remote\-[a-f0-9]+")
--- a/lib/charms/istio_beacon_k8s/v0/service_mesh.py
+++ b/lib/charms/istio_beacon_k8s/v0/service_mesh.py
@@ -133,7 +133,7 @@ class ServiceMeshConsumer(Object):
                         {
                             "app_name": relation.app.name,
                             "namespace": self._my_namespace(),
-                            "endpoints": policy.endpoints,
+                            "endpoints": [endpoint.model_dump() for endpoint in policy.endpoints],
                             "service": policy.service,
                         }
                     )
--- a/lib/charms/istio_beacon_k8s/v0/service_mesh.py
+++ b/lib/charms/istio_beacon_k8s/v0/service_mesh.py
@@ -13,12 +13,15 @@ To add service mesh support to your charm, instantiate a ServiceMeshConsumer obj
 from charms.istio_beacon_k8s.v0.service_mesh import Policy, ServiceMeshConsumer
 
 ...
-self._mesh = ServiceMeshConsumer(
-    self,
-    policies=[
-        Policy(relation="logging", endpoints=[f"*:{HTTP_LISTEN_PORT}"]),
-    ],
-)
+try:
+    self._mesh = ServiceMeshConsumer(
+        self,
+        policies=[
+            Policy(relation="logging", endpoints=[f"*:{HTTP_LISTEN_PORT}"]),
+        ],
+    )
+except ops.TooManyRelatedAppsError as e:
+    self.unit.status = BlockedStatus(e)
 ```
 """
 
@@ -63,7 +66,7 @@ class ServiceMeshConsumer(Object):
         """
         super().__init__(charm, mesh_relation_name)
         self._charm = charm
-        self._relations = self._charm.model.relations[mesh_relation_name]
+        self._relation = self._charm.model.get_relation[mesh_relation_name]
         self._policies = policies or []
         self.framework.observe(
             self._charm.on[mesh_relation_name].relation_created, self._relations_changed
@@ -109,8 +112,7 @@ class ServiceMeshConsumer(Object):
             "model": self._my_namespace(),
             "policies": policies,
         }
-        for rel in self._relations:
-            rel.data[self._charm.app]["mesh_data"] = json.dumps(mesh_rel_data)
+        self._relation.data[self._charm.app]["mesh_data"] = json.dumps(mesh_rel_data)
 
     def _my_namespace(self):
         """Return the namespace of the running charm."""
--- a/lib/charms/istio_beacon_k8s/v0/service_mesh.py
+++ b/lib/charms/istio_beacon_k8s/v0/service_mesh.py
@@ -48,7 +48,7 @@ class Policy(pydantic.BaseModel):
 
 
 class ServiceMeshConsumer(Object):
-    """Class for managing the consumer side of the service_mesh relation interface."""
+    """Class used for joining a service mesh."""
 
     def __init__(
         self,
--- a/lib/charms/istio_beacon_k8s/v0/service_mesh.py
+++ b/lib/charms/istio_beacon_k8s/v0/service_mesh.py
@@ -71,6 +71,7 @@ class Policy(pydantic.BaseModel):
 
     relation: str
     endpoints: List[Endpoint]
+    service: Optional[str]
 
 
 class ServiceMeshConsumer(Object):
@@ -133,6 +134,7 @@ class ServiceMeshConsumer(Object):
                             "app_name": relation.app.name,
                             "namespace": self._my_namespace(),
                             "endpoints": policy.endpoints,
+                            "service": policy.service,
                         }
                     )
         mesh_rel_data = {
--- a/lib/charms/istio_beacon_k8s/v0/service_mesh.py
+++ b/lib/charms/istio_beacon_k8s/v0/service_mesh.py
@@ -25,6 +25,7 @@ except ops.TooManyRelatedAppsError as e:
 ```
 """
 
+import enum
 import json
 import logging
 import re
@@ -37,14 +38,33 @@ LIBID = "3f40cb7e3569454a92ac2541c5ca0a0c"  # Never change this
 LIBAPI = 0
 LIBPATCH = 1
 
+PYDEPS = ["pydantic"]
+
 logger = logging.getLogger(__name__)
+
+
+class Method(str, enum.Enum):
+    """HTTP method."""
+
+    connect = "CONNECT"
+    delete = "DELETE"
+    get = "GET"
+    head = "HEAD"
+    options = "OPTIONS"
+    patch = "PATCH"
+    post = "POST"
+    put = "PUT"
+    trace = "TRACE"
 
 
 class Policy(pydantic.BaseModel):
     """Data type for holding a service mesh policy."""
 
     relation: str
-    endpoints: List[str]
+    hosts: List[str]
+    ports: List[int]
+    methods: List[Method]
+    paths: List[str]
 
 
 class ServiceMeshConsumer(Object):
--- a/lib/charms/istio_beacon_k8s/v0/service_mesh.py
+++ b/lib/charms/istio_beacon_k8s/v0/service_mesh.py
@@ -17,7 +17,18 @@ try:
     self._mesh = ServiceMeshConsumer(
         self,
         policies=[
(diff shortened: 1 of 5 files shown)


lib/charms/istio_beacon_k8s/v0/service_mesh.py
```diff
@@ -0,0 +1,119 @@
+# Copyright 2024 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""#Service Mesh Library.
```
ca-scribner, line 4: I see the `#` formatter in a few libs.  Do these descriptions get published in charmhub?

lib/charms/istio_beacon_k8s/v0/service_mesh.py
```diff
@@ -0,0 +1,119 @@
...
+
+The service mesh library is used to facilitate adding you charmed application to a service mesh.
+The library leverages the `service_mesh` interface and the `cross_model_mesh` interface
+
```
ca-scribner, line 8: iiuc the `cross_model_mesh` isn't needed anymore, but if I'm wrong reject this.

```suggestion
The service mesh library is used to facilitate adding your charmed application to a service mesh.
The library leverages the `service_mesh` interface.

```

lib/charms/istio_beacon_k8s/v0/service_mesh.py
```diff
@@ -0,0 +1,119 @@
...
+                            "app_name": relation.app.name,
+                            "namespace": self._my_namespace(),
+                            "endpoints": policy.endpoints,
+                        }
+                    )
+        mesh_rel_data = {
+            "app_name": self._charm.app.name,
+            "model": self._my_namespace(),
+            "policies": policies,
+        }
```
ca-scribner, line 111: Trying to think of the different use cases for this relation.  It could enable communication the following cases:
1. let `other_app`'s charm/sidecar talk to `this_app`'s charm/sidecar
1. let `other_app`'s kubectl-deployed workload talk to `this_app`'s charm/sidecar
1. let `other_app`'s charm/sidecar talk to `this_app`'s kubectl-deployed workload
1. let `other_app`'s kubectl-deployed workload talk to `this_app`'s kubectl-deployed workload

which of these cases do we want to solve?  I think we definitely have (1), but I'm less sure about (4).  I think we talked about a git-sync solution for the really complex stuff - do some of these cases fall into the git-sync solution scope?

lib/charms/istio_beacon_k8s/v0/service_mesh.py
dstathis: Yes in the "Libraries" tab.

lib/charms/istio_beacon_k8s/v0/service_mesh.py
dstathis: Yeah we still need cross_model_mesh for sure.

lib/charms/istio_beacon_k8s/v0/service_mesh.py
```diff
@@ -0,0 +1,119 @@
...
+The library leverages the `service_mesh` interface and the `cross_model_mesh` interface
+
+To add service mesh support to your charm, instantiate a ServiceMeshConsumer object in the
+`__init__` method of your charm:
+
+```
+from charms.istio_beacon_k8s.v0.service_mesh import IsCrossModelError, Policy, ServiceMeshConsumer
+
+...
+self._mesh = ServiceMeshConsumer(
```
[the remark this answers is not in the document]
dstathis: Sure opening an issue now.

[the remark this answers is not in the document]
dstathis: #4

lib/charms/istio_beacon_k8s/v0/service_mesh.py
```diff
@@ -0,0 +1,119 @@
...
+LIBPATCH = 1
+
+logger = logging.getLogger(__name__)
+
+
+class Policy(pydantic.BaseModel):
+    """Data type for holding a service mesh policy."""
+
+    relation: str
+    endpoints: List[str]
```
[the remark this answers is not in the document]
dstathis: I kind of ignored the "what is an endpoint" question here. Since we want to make this mesh agnostic, it might not map exactly to everything Istio has to offer. I will check out how other meshes (at least Cilium) specify destinations, and make a decision as to what an endpoint should be. If it does turn out to be a string, I will at least document it in the class docstring.

lib/charms/istio_beacon_k8s/v0/service_mesh.py
```diff
@@ -0,0 +1,119 @@
...
+                if cmr_matcher.fullmatch(relation.app.name):
+                    logger.debug(
+                        f"Cross model relation found: {relation.name}. Currently not implemented. Skipping."
+                    )
+                else:
+                    logger.debug(f"Found relation: {relation.name}. Creating policy.")
+                    policies.append(
+                        {
+                            "app_name": relation.app.name,
+                            "namespace": self._my_namespace(),
```
IbraAoad, line 103: afaik and from looking at how the [`AuthPolicy`](https://istio.io/latest/docs/reference/config/security/authorization-policy/) is structured, we mainly have 3 sections

1. targetRefs
this specifies where to allow traffic to go to, in this case we would say allow charm A to talk to me 
  -  Workloadselector
    we can use this instead of target refs to select destination by label instead of explicitly specifying namespace and service name (I think if we need this if charm A is setting up a policy to allow traffic from charm B to go to charm C?)

2. From
source principals: the name of the service account the traffic will come from (in juju this matches app name and we should also specify the namespace if the policy and the source app aren't on the same namespace)
source namespace: we can say allow all traffic that comes from this namespace

3. To
the L7 policies, basically which port or prefix


now based on the above think of this scenario, (Loki deployed on model A and Zookeeper deployed on model B) loki wants to say allow zookeeper to talk to me over `:80`, loki would instantiate the `ServiceMeshConsumer` with the name of the `logging` relation and the `endpoint`, how will we figure out the namespace for zookeeper? cause if we don't the policy will assume that zookeeper is sitting in the same namespace as loki.

lib/charms/istio_beacon_k8s/v0/service_mesh.py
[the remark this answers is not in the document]
ca-scribner: this doesn't have to be solved in this PR either - it isn't a big deal to change until something actually starts using this interface since we can just change it without backward compatibility or revisioning.  If there's a small chance that we want the flexibility though, we might as well just make an `Endpoint` object even if its data starts just as `{string_endpoint: str}`

lib/charms/istio_beacon_k8s/v0/service_mesh.py
```diff
@@ -0,0 +1,119 @@
...
+                            "endpoints": policy.endpoints,
+                        }
+                    )
+        mesh_rel_data = {
+            "app_name": self._charm.app.name,
+            "model": self._my_namespace(),
```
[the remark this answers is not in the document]
dstathis: I have some security concerns about the custom resource selector. I would prefer to simply ask charm authors to add the label "app.kubernetes.io/name=<app_name>" to any custom resources. We should discuss this. Also, I still need to confirm that manually adding that label will not mess with juju.

lib/charms/istio_beacon_k8s/v0/service_mesh.py
dstathis: If they are related cross model they will need to use the additional mesh_cross_model relation which communicates that information.

dstathis: That support will be in the next iteration of the lib.

lib/charms/istio_beacon_k8s/v0/service_mesh.py
dstathis: We are trying to solve all 4. The idea being that if a charm wants to be on the mesh with custom workloads, then need to add the appropriate app name label to anything they kubectl-deploy.

lib/charms/istio_beacon_k8s/v0/service_mesh.py
```diff
@@ -0,0 +1,119 @@
...
+            "policies": policies,
+        }
+        for rel in self._relations:
+            rel.data[self._charm.app]["mesh_data"] = json.dumps(mesh_rel_data)
```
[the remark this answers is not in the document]
dstathis: Agreed. Updated.

lib/charms/istio_beacon_k8s/v0/service_mesh.py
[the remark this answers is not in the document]
dstathis: 3 is still correct. Beacon adding the whole model to the mesh will be an optional config. Relating to beacon will also join you to the mesh.

lib/charms/istio_beacon_k8s/v0/service_mesh.py
```diff
@@ -0,0 +1,119 @@
...
+
+class Policy(pydantic.BaseModel):
+    """Data type for holding a service mesh policy."""
+
+    relation: str
+    endpoints: List[str]
+
+
+class ServiceMeshConsumer(Object):
+    """Class for managing the consumer side of the service_mesh relation interface."""
```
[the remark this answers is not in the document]
dstathis: I agree.

lib/charms/istio_beacon_k8s/v0/service_mesh.py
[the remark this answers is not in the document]
ca-scribner: Right, I remember you mentioning this, and agreed on the security concerns.  

What if a charm deploys two workloads, can we distinguish between them?  This might be too restrictive.  

How about us define a key for our purpose (rather than `app.kubernetes.io/name`), and require the `value` be prefixed by the application name.  Ex: `canonical.com/istio-beacon-selector: APPNAME-WHATEVER`, where the related app can put whatever it wants in `WHATEVER`.  I think that would lock things down enough, but still be flexible.  wdyt?

lib/charms/istio_beacon_k8s/v0/service_mesh.py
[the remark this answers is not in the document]
ca-scribner: Are all 4 cases in scope for this PR?  If yes, could you give an example of what two apps would be sending beacon to make it work?  It isn't all coming together in my head

lib/charms/istio_beacon_k8s/v0/service_mesh.py
ca-scribner: Are all 4 cases in scope for this PR?  If yes, could you give an example of what two apps would be sending beacon to make it work?  It isn't all coming together in my head

lib/charms/istio_beacon_k8s/v0/service_mesh.py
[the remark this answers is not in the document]
ca-scribner: k, then maybe what I'm unsure about is what is in scope for this PR.  Could you outline what is/isn't implemented here?  I'm not sure if I'm finding things that are missing, or just things that are coming in future PRs.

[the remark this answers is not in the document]
dstathis: In scope for this PR: Using the constructor and relations to generate the relation data required to allow a service mesh to generate access policies.

In scope for the next PR: Add the cross_model_mesh interface support to allow for cross model relations.

Unrelated to the lib: Everything else. (model join for example has nothing to do with relations).

lib/charms/istio_beacon_k8s/v0/service_mesh.py
```diff
@@ -0,0 +1,119 @@
...
+        allow communication.
+        """
+        logger.debug("Updating service mesh policies.")
+        policies = []
+        cmr_matcher = re.compile(r"remote\-[a-f0-9]+")
+        for policy in self._policies:
+            for relation in self._charm.model.relations[policy.relation]:
+                if cmr_matcher.fullmatch(relation.app.name):
+                    logger.debug(
+                        f"Cross model relation found: {relation.name}. Currently not implemented. Skipping."
```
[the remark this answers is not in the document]
dstathis: Done.

lib/charms/istio_beacon_k8s/v0/service_mesh.py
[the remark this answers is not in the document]
dstathis: Okay so I decided to break up endpoints in to four lists: hosts, ports, methods, and paths.

What do you think about this approach?

lib/charms/istio_beacon_k8s/v0/service_mesh.py
ca-scribner: Ok, so then let's sketch out the `AuthorizationPolicy` this all eventually renders to for the "A talks to B at `:9090/bar`" example:

```yaml
apiVersion: security.istio.io/v1
kind: AuthorizationPolicy
metadata:
  name: let-a-talk-to-b
  namespace: B-NAMESPACE
spec:
  # One of either selector or targetRefs (not both)
  selector:
    matchLables:
      B-SELECTOR-KEYS: B-SELECTOR-VALUES
  targetRefs:  # This is new, from the Gateway API, and might work better for us.  Can point to Gateway/Service/...
  - kind: Service
    group: ""
    name: B-SERVICE  # Where this is a service of the waypoint proxy (I think?  or a service backed by a waypoint proxy?)
  action: ALLOW
  rules:
  - from:
    - source:
        principals:
        - A-SERVICE-ACCOUNT  # note this includes the namespace and service account name
    to:
    - operation:  # Where on B-SERVICE we can communicate with
      - paths: B-PATHS
        ports: B-PORTS
```

So for the data we need:
* `A-SERVICE-ACCOUNT`: 
  * sidecar workload :heavy_check_mark: : beacon can infer this from juju (service account will be `appname`)
  * kubectl workload :question:: this is an arbitrary name and we need to send it via the relation
* `B-SERVICE`:
  * sidecar workload :heavy_check_mark:: beacon can infer this from juju (service will be `appname`)
  * kubectl workload :question:: this is an arbitrary name and we need to send it via the relation
* `B-SELECTOR`:
  * sidecar workload :heavy_check_mark:: beacon can infer this from juju's default labels
  * kubectl workload :question:: I think we need something more here ([this comment](https://github.com/canonical/istio-beacon-k8s-operator/pull/3#discussion_r1687087130))
* `B-PATHS`/`B-PORTS` :heavy_check_mark:: from `Policy` `Endpoints`

I think we're missing some of the kubectl parts above.

Also, I didn't realize istio extended the `AuthorizationPolicy` to include `targetRefs`.  Those might make more sense for us to use than the `WorkloadSelector`, as they can point to a `Service`

lib/charms/istio_beacon_k8s/v0/service_mesh.py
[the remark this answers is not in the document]
dstathis: This seems doable but would add a significant level of complexity. I think in the vast majority of cases specifying port, path, etc. should be enough to differentiate between workloads. WDYT?

lib/charms/istio_beacon_k8s/v0/service_mesh.py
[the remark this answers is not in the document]
dstathis: I used the former. I already implemented the change so you can look at the code.

lib/charms/istio_beacon_k8s/v0/service_mesh.py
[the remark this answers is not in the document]
dstathis: Sorry you are right it also joins charms to the mesh that relate over service_mesh.

lib/charms/istio_beacon_k8s/v0/service_mesh.py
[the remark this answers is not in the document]
ca-scribner: If the same workload selector applies to multiple workloads, then they'll both get the same policy.  So if `X` talks to `A` at `/foo` and `Y` talks to `B` at `/bar`, a common workload selector will resolve to `both X and Y can talk to both A/foo and B/bar`.  Often that'll be ok, but there's probably good cases when that is a security issue.  Whether that's an issue for us is a good question ...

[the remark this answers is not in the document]
ca-scribner: `A` and `B` in this scenario will already need to have unique workload selectors for their respective service objects (otherwise `X` and `Y` could never actually reach them in the first place).  I wonder if, based on how AuthorizationPolicies now support applying policies against a Service (instead of just a WorkloadSelector like they used to) ([mentioned here](https://github.com/canonical/istio-beacon-k8s-operator/pull/3#discussion_r1688695038)), we should just ignore `WorkloadSelectors` here and instead pass `Service` names.

lib/charms/istio_beacon_k8s/v0/service_mesh.py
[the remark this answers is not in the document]
ca-scribner: :+1: lgtm, we can catch it in a future pr then

lib/charms/istio_beacon_k8s/v0/service_mesh.py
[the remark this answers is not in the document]
ca-scribner: sorry, I may be really confused.  I only see this:

```python
class Policy(pydantic.BaseModel):
    """Data type for holding a service mesh policy."""

    relation: str
    endpoints: List[str]
```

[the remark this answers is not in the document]
dstathis: Yup my bad. Pushed now.

lib/charms/istio_beacon_k8s/v0/service_mesh.py
ca-scribner: We discussed and decided to focus only on the sidecar-deployed workloads for now.  We'll tackle the kubectl-deployed things in future.

lib/charms/istio_beacon_k8s/v0/service_mesh.py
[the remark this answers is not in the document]
ca-scribner: As discussed, we'll switch this to use the Service instead of WorkloadSelector, which makes things a bit easier.

lib/charms/istio_beacon_k8s/v0/service_mesh.py
sed-i, line 1: You probably already considered this, but just in case:
Since interface names are (implicitly) unique in the ecosystem, is this intended to be "service_mesh", or should it be "istio_service_mesh"?
(The relation name can obviously still be "service-mesh".)

lib/charms/istio_beacon_k8s/v0/service_mesh.py
```diff
@@ -0,0 +1,149 @@
+# Copyright 2024 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""#Service Mesh Library.
+
+The service mesh library is used to facilitate adding you charmed application to a service mesh.
```
sed-i, line 6:
```suggestion
The service mesh library is used to facilitate adding your charmed application to a service mesh.
```

lib/charms/istio_beacon_k8s/v0/service_mesh.py
```diff
@@ -0,0 +1,149 @@
...
+The library leverages the `service_mesh` interface and the `cross_model_mesh` interface
```
sed-i, line 7:
```suggestion
The library leverages the `service_mesh` and `cross_model_mesh` interfaces.
```

lib/charms/istio_beacon_k8s/v0/service_mesh.py
```diff
@@ -0,0 +1,149 @@
...
+
+To add service mesh support to your charm, instantiate a ServiceMeshConsumer object in the
```
sed-i, line 9: Not familiar with the tech so just checking: Is there a server-client situation here in the context of this relation? Because the server should be the provider, and it seems like "your charm" is likely the server.

lib/charms/istio_beacon_k8s/v0/service_mesh.py
```diff
@@ -0,0 +1,149 @@
...
+
+```
+from charms.istio_beacon_k8s.v0.service_mesh import Policy, ServiceMeshConsumer
+
+...
+try:
+    self._mesh = ServiceMeshConsumer(
+        self,
+        policies=[
+            Policy(relation="logging", endpoints=[f"*:{HTTP_LISTEN_PORT}"]),
```
sed-i, line 20: "Policy" is a bit ambiguous.
Compare to urlparse:
```python
from urllib.parse import urlparse
```
The module name `parse` isn't prefixed with "url", but the func "parse" is, to avoid potential name collisions.

I.e. should Policy be renamed to e.g. "ServiceMeshPolicy"?

lib/charms/istio_beacon_k8s/v0/service_mesh.py
```diff
@@ -0,0 +1,151 @@
...
+    trace = "TRACE"
+
+
+class Endpoint(pydantic.BaseModel):
+    """Data type for a policy endpoint."""
+
+    hosts: List[str]
+    ports: List[int]
+    methods: List[Method]
+    paths: List[str]
```
sed-i, line 66: Add a comment for each? No clue what they're for.
Same comment for the Policy BaseModel.

lib/charms/istio_beacon_k8s/v0/service_mesh.py
```diff
@@ -0,0 +1,151 @@
...
+
+        Args:
+            charm: The charm instantiating this object.
+            mesh_relation_name: The relation name as defined in metadata.yaml or charmcraft.yaml
+                for the relation which used the service_mesh interface.
+            policies: List of access policies this charm supports.
+        """
+        super().__init__(charm, mesh_relation_name)
+        self._charm = charm
+        self._relation = self._charm.model.get_relation(mesh_relation_name)
```
sed-i, line 96: This assumes `limit: 1`, but no limit is specified in metadata.

lib/charms/istio_beacon_k8s/v0/service_mesh.py
```diff
@@ -0,0 +1,151 @@
...
+        """Update the service mesh.
+
+        Gathers information from all relations of the charm and updates the mesh appropriately to
+        allow communication.
+        """
+        if self._relation is None:
+            return
+        logger.debug("Updating service mesh policies.")
+        policies = []
+        cmr_matcher = re.compile(r"remote\-[a-f0-9]+")
```
sed-i, line 123: Not sure why this is here for.
- The `remote-` prefix is not guaranteed because offerer/consumer are interchangeable.
- How would implementing the cmr iface change this?

lib/charms/istio_beacon_k8s/v0/service_mesh.py
```diff
@@ -0,0 +1,151 @@
...
+        for policy in self._policies:
+            for relation in self._charm.model.relations[policy.relation]:
+                if cmr_matcher.fullmatch(relation.app.name):
+                    logger.debug(
+                        f"Cross model relation found: {relation.name}. Currently not implemented. Skipping."
+                    )
+                else:
+                    logger.debug(f"Found relation: {relation.name}. Creating policy.")
```
sed-i, line 131: Let's make this log line more useful by including some details from the policy.

lib/charms/istio_beacon_k8s/v0/service_mesh.py
dstathis: It is supposed to be a generic so `service_mesh`.

lib/charms/istio_beacon_k8s/v0/service_mesh.py
dstathis: Your charm in this situation is joining the mesh. Istio beacon is "providing" that functionality.

lib/charms/istio_beacon_k8s/v0/service_mesh.py
dstathis: It will always be `charms.istio-beacon.v0.service_mesh.Policy` which will never collide with anything.

lib/charms/istio_beacon_k8s/v0/service_mesh.py
dstathis: The `limit: 1` should be on the consumer side not the provider.

lib/charms/istio_beacon_k8s/v0/service_mesh.py
dstathis: I realize this probably does not work correctly yet but I plan on fixing it with a CMR support PR.

lib/charms/istio_beacon_k8s/v0/service_mesh.py
dstathis: I think that would be a bit to verbose and could maybe be a security vulnerability as well.

lib/charms/istio_beacon_k8s/v0/service_mesh.py
dstathis: These are parts of url requests. I'm not sure how to describe them really other than just repeating the names.

PR #5, opened 2024-08-07
Istio Beacon - Waypoint

IbraAoad:
## Context
Adding an L7 Gateway(waypoint) per beacon  

## Scope
This PR will be limited to the below
- Deploys a waypoint `on start` that sits on the same namespace as the beacon and can manage the traffic across all namespaces.
- Have a config option `model-on-mesh` that will put the beacon's model on the mesh along with configuring it to use the waypoint it deployed.

## Work to do
- Figure out a way to put a single charm on the mesh and similarly configure a single charm/workload to use a waypoint instead of the whole model #6

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -5,7 +5,11 @@ on:
     branches:
       - main
 
-# TODO: Remove when #8 closes
+# The following configuration is temporarily pinned to this branch to enable Juju to restart hooks, 
+# which are disabled by default (https://github.com/charmed-kubernetes/actions-operator/blob/5c6377ed695d52b8a1693f07b7d641e245269123/src/bootstrap/index.ts#L205). 
+# Note: This causes integration tests to fail as charms lose connection to the Juju controller 
+# when added to the mesh, regaining connection only upon hook retry.
+# TODO: Revert to main branch after #8 is resolved.
 jobs:
   pull-request:
     name: PR
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -5,6 +5,7 @@ on:
     branches:
       - main
 
+# TODO: Remove when #8 closes
 jobs:
   pull-request:
     name: PR
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -8,5 +8,5 @@ on:
 jobs:
   pull-request:
     name: PR
-    uses: canonical/observability/.github/workflows/charm-pull-request.yaml@main
+    uses: canonical/observability/.github/workflows/charm-pull-request.yaml@istio-test
     secrets: inherit
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -60,4 +60,5 @@ config:
       type: boolean
       default: false
       description: >
-        Add the whole model on the mesh and connect it to the waypoint
+        Add this charm's model to the service mesh. 
+        All charms in this model will automatically be added to the mesh.
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -43,3 +43,21 @@ parts:
       # Pydantic is not actually used by the lib but charmcraft install pydeps from libs even if
       # they are not used.
       - pydantic>2.0
+    plugin: charm
+    build-packages: [git]
+
+config:
+  options:
+    ready-timeout:
+      type: int
+      default: 100
+      description: >
+        The maximum time (in seconds) to wait for the waypoint deployment to be 
+        ready. This applies specifically to the deployment created for the Istio 
+        waypoint controller. If the deployment does not become ready within this time, 
+        charm will go into blocked state.
+    model-on-mesh:
+      type: bool
+      default: False
+      description: >
+        Add the whole model on the mesh and connect it to the waypoint
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -55,7 +55,7 @@ config:
         The maximum time (in seconds) to wait for the waypoint deployment to be 
         ready. This applies specifically to the deployment created for the Istio 
         waypoint controller. If the deployment does not become ready within this time, 
-        charm will go into blocked state.
+        charm will go into error state.
     model-on-mesh:
       type: boolean
       default: false
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -57,7 +57,7 @@ config:
         waypoint controller. If the deployment does not become ready within this time, 
         charm will go into blocked state.
     model-on-mesh:
-      type: bool
-      default: False
+      type: boolean
+      default: false
       description: >
         Add the whole model on the mesh and connect it to the waypoint
--- a/src/charm.py
+++ b/src/charm.py
@@ -171,8 +171,8 @@ class IstioBeaconCharm(ops.CharmBase):
         try:
             namespace = self.lightkube_client.get(Namespace, self.model.name)
         except ApiError as e:
-                logger.error(f"Error checking namespace labels {e}")
-                return
+            logger.error(f"Error checking namespace labels {e}")
+            return
 
         if add_labels:
             labels_to_add = {
@@ -196,7 +196,7 @@ class IstioBeaconCharm(ops.CharmBase):
         try:
             self.lightkube_client.patch(Namespace, self.model.name, namespace)
         except ApiError as e:
-                logger.error(f"Error patching istio labels {e}")  
+            logger.error(f"Error patching istio labels {e}")
 
 
 if __name__ == "__main__":
--- a/src/charm.py
+++ b/src/charm.py
@@ -16,8 +16,9 @@ from lightkube.models.meta_v1 import ObjectMeta
 from lightkube.resources.apps_v1 import Deployment
 from lightkube.resources.core_v1 import Namespace
 from lightkube_extensions.batch import KubernetesResourceManager, create_charm_default_labels
-from models import AllowedRoutes, IstioWaypointResource, IstioWaypointSpec, Listener, Metadata
 from ops.model import ActiveStatus, BlockedStatus, MaintenanceStatus
+
+from models import AllowedRoutes, IstioWaypointResource, IstioWaypointSpec, Listener, Metadata
 
 logger = logging.getLogger(__name__)
 
@@ -114,6 +115,9 @@ class IstioBeaconCharm(ops.CharmBase):
         return True
 
     def _sync_all_resources(self):
+        if not self.unit.is_leader():
+            self.unit.status = BlockedStatus("Waypoint can only be provided on the leader unit.")
+            return
         self.unit.status = MaintenanceStatus("Validating waypoint readiness")
         self._sync_waypoint_resources()
         if not self._is_ready():
@@ -144,14 +148,12 @@ class IstioBeaconCharm(ops.CharmBase):
         )
         gateway_resource = RESOURCE_TYPES["Gateway"]
         return gateway_resource(
-            metadata=ObjectMeta.from_dict(gateway.metadata.dict()),
-            spec=gateway.spec.dict(),
+            metadata=ObjectMeta.from_dict(gateway.metadata.model_dump()),
+            spec=gateway.spec.model_dump(),
         )
 
     def _sync_waypoint_resources(self):
         resources_list = []
-        if not self.unit.is_leader():
-            raise RuntimeError("Waypoint can only be provided on the leader unit.")
         krm = self._get_waypoint_resource_manager()
         resource_to_append = self._construct_waypoint()
         resources_list.append(resource_to_append)
--- a/src/charm.py
+++ b/src/charm.py
@@ -60,7 +60,7 @@ class IstioBeaconCharm(ops.CharmBase):
 
     def _on_remove(self, _):
         """Event handler for remove."""
-        self._update_labels(False)
+        self._remove_labels()
         krm = self._get_waypoint_resource_manager()
         krm.delete()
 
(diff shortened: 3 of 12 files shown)


charmcraft.yaml
```diff
@@ -43,3 +43,21 @@ parts:
...
+config:
+  options:
+    ready-timeout:
+      type: int
+      default: 100
+      description: >
+        The maximum time (in seconds) to wait for the waypoint deployment to be 
+        ready. This applies specifically to the deployment created for the Istio 
+        waypoint controller. If the deployment does not become ready within this time, 
+        charm will go into blocked state.
```
ca-scribner, line 58: I'm not sure this should be user-facing config.  Are there cases where we can't set a reasonable default?

charmcraft.yaml
```diff
@@ -43,3 +43,21 @@ parts:
...
+    model-on-mesh:
+      type: boolean
+      default: false
+      description: >
+        Add the whole model on the mesh and connect it to the waypoint
```
ca-scribner, line 63: Just a suggestion.
```suggestion
        Add this charm's model to the service mesh. 
 All charms in this model will automatically be added to the mesh. 

```

Not sure if you need to mention the waypoint here.  I think somewhere in readme or elsewhere we'd describe what this charm does (incl deploy a waypoint) and then this becomes redundant here

src/charm.py
```diff
@@ -6,38 +6,197 @@
...
+        resource_to_append = self._construct_waypoint()
+        resources_list.append(resource_to_append)
+        krm.reconcile(resources_list)
+
+        if self.config["model-on-mesh"]:
+            self._update_labels()
+        else:
+            self._update_labels(False)
+
+    def _update_labels(self, add_labels=True):
```
ca-scribner, line 165: I'm torn here.  I like how we reduce duplicate code by combing `add_label` and `remove_label` into `update_label`, but it feels odd to call `_update_labels(add_labels=False)` to **remove** labels.  I dont have a suggestion on how to rename this combined function though

I lean toward splitting add and remove into separate functions, then sharing code between them with helpers

src/charm.py
```diff
@@ -6,38 +6,197 @@
...
+
+            if (
+                namespace.metadata
+                and namespace.metadata.labels
+                and (
+                    "istio.io/use-waypoint" in namespace.metadata.labels
+                    or "istio.io/dataplane-mode" in namespace.metadata.labels
+                )
+            ):
+                namespace.metadata.labels.update(labels_to_remove)
```
ca-scribner, line 195: we can simplify this a little by (I left out the error checking as I'd probably make a mistake here, but you'd have to add it back):
```suggestion
        labels = {
            "istio.io/use-waypoint": None,
            "istio.io/dataplane-mode": None,
        }
        if add_labels:
            labels["istio.io/use-waypoint"] = f"{self.app.name}-{self.model.name}-waypoint"
            labels["istio.io/dataplane-mode"] = "ambient"
        namespace.metadata.labels.update(labels)
```

src/charm.py
```diff
@@ -6,38 +6,197 @@
...
+
+        if add_labels:
+            labels_to_add = {
+                "istio.io/use-waypoint": f"{self.app.name}-{self.model.name}-waypoint",
+                "istio.io/dataplane-mode": "ambient",
+            }
+            if namespace.metadata and namespace.metadata.labels:
+                namespace.metadata.labels.update(labels_to_add)
+        else:
+            labels_to_remove = {"istio.io/use-waypoint": None, "istio.io/dataplane-mode": None}
```
ca-scribner, line 185: as is, this is going to prevent the user from setting their own waypoint labels on a namespace with beacon in it.  Ex: 
```
juju add-model test
kubectl label namespace istio.io/use-waypoint=mywaypoint
juju deploy beacon
# label is removed from namespace
```

Is that ok?  I'm not sure, but if yes we should clearly state that in docs.  

If no, we can get around it by setting an additional label like `istio.io/use-waypoint/managed-by=f"{self.app.name}-{self.model.name}"` and we only edit/remove waypoint labels if we see that additional label

src/charm.py
```diff
@@ -6,38 +6,197 @@
...
+                        protocol="HBONE",
+                        allowedRoutes=AllowedRoutes(namespaces={"from": "All"}),
+                    )
+                ],
+            ),
+        )
+        gateway_resource = RESOURCE_TYPES["Gateway"]
+        return gateway_resource(
+            metadata=ObjectMeta.from_dict(gateway.metadata.dict()),
+            spec=gateway.spec.dict(),
```
ca-scribner, line 148: dict() is deprecated
```suggestion
            spec=gateway.spec.model_dump(),
```

src/charm.py
```diff
@@ -6,38 +6,197 @@
...
+                        port=15008,
+                        protocol="HBONE",
+                        allowedRoutes=AllowedRoutes(namespaces={"from": "All"}),
+                    )
+                ],
+            ),
+        )
+        gateway_resource = RESOURCE_TYPES["Gateway"]
+        return gateway_resource(
+            metadata=ObjectMeta.from_dict(gateway.metadata.dict()),
```
ca-scribner, line 147: dict() is deprecated
```suggestion
            metadata=ObjectMeta.from_dict(gateway.metadata.model_dump()),
```

src/charm.py
```diff
@@ -6,38 +6,197 @@
...
+            spec=gateway.spec.dict(),
+        )
+
+    def _sync_waypoint_resources(self):
+        resources_list = []
+        if not self.unit.is_leader():
+            raise RuntimeError("Waypoint can only be provided on the leader unit.")
```
ca-scribner, line 154: Can we change this to a BlockedStatus rather than raising an error?  Error will tell juju to keep retrying this event until it succeeds, so a non-leader unit will keep retrying the same event over and over and hitting this error.

I **think** if it eventually was elected leader it would recover, but I'm not certain of that...  either way, Blocked conveys enough of a message imo

ca-scribner (commented):
a few comments but generally lgtm

src/charm.py
IbraAoad: `istio.io/use-waypoint/managed-by` seems reasonable.

charmcraft.yaml
IbraAoad: This was a recommendation from @dstathis and @simskij, mainly figuring out a reasonable default for all k8s distros is going to be very volatile, so we thought we would leave it at 100 and give the user the ability to increase based on his reqs.

src/charm.py
IbraAoad: had to update the label to be `istio.io.waypoint-managed-by` because of the below error,

`invalid: metadata.labels: Invalid value: "istio.io/use-waypoint/managed-by": a qualified name must consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyName',  or 'my.name',  or '123-abc', regex used for validation is '([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]') with an optional DNS subdomain prefix and '/' (e.g. 'example.com/MyName')`

.github/workflows/pull-request.yaml
```diff
@@ -8,5 +8,5 @@ on:
 jobs:
   pull-request:
     name: PR
-    uses: canonical/observability/.github/workflows/charm-pull-request.yaml@main
+    uses: canonical/observability/.github/workflows/charm-pull-request.yaml@istio-test
```
IbraAoad, line 11: This is pinned to this branch for now to allow juju to restart hooks as they are disabled by [default ](https://github.com/charmed-kubernetes/actions-operator/blob/5c6377ed695d52b8a1693f07b7d641e245269123/src/bootstrap/index.ts#L205), this breaks itests as charms will lose connections to the juju controller once added to the mesh and will regain the connection on the hook retry

ca-scribner: If this is going to stay in, can you add a comment in the code?  Could just open an issue then add a `TODO: Remove when http://...issues/123 closes` comment

src/charm.py
ca-scribner: oh yep, rules are [here](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set)

I have suggestions for modifications, but I'll open that separately

ca-scribner: oh yep, rules are [here](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set)

I have suggestions for modifications, but I'll open that separately

src/charm.py
```diff
@@ -6,38 +6,231 @@
...
+            return False
+        return True
+
+    def _sync_all_resources(self):
+        if not self.unit.is_leader():
+            self.unit.status = BlockedStatus("Waypoint can only be provided on the leader unit.")
+            return
+        self.unit.status = MaintenanceStatus("Validating waypoint readiness")
+        self._sync_waypoint_resources()
+        if not self._is_ready():
```
ca-scribner, line 123: We should add a log message or maybe even a MaintenanceStatus to say that we're checking whether everything we deployed is up.  Use your judgement on where/what, but atm the charm might sit seemingly doing nothing for 15 minutes.  Maybe a log message saying it is checking, and an update message every time it checks and fails?

src/charm.py
```diff
@@ -6,38 +6,231 @@
...
+        try:
+            self.lightkube_client.patch(Namespace, self.model.name, namespace)
+        except ApiError as e:
+            logger.error(f"Error patching namespace: {e}")
+
+    def _add_labels(self):
+        """Add specific labels to the namespace."""
+        namespace = self._get_namespace()
+        if not namespace:
+            return
```
ca-scribner, line 186: Is there a case where this is not an error?  If not, we should raise here.

src/charm.py
```diff
@@ -6,38 +6,231 @@
...
+                )
+                return
+
+            labels_to_add = {
+                "istio.io/use-waypoint": f"{self.app.name}-{self.model.name}-waypoint",
+                "istio.io/dataplane-mode": "ambient",
+                "istio.io.waypoint-managed-by": f"{self.app.name}-{self.model.name}",
+            }
+
+            namespace.metadata.labels.update(labels_to_add)
```
ca-scribner, line 207: I think this will fail if we try to `_add_labels` to a `namespace` that has no labels because I think `labels` will be None not an empty dict, and then `namespace.metadata.labels.update()` will try to `.update()` on None.  I'm not 100% sure though

What we're definitely missing here are some unit tests.  There's lots of logic here and its nicely encapsulated for testing - let's add some unit tests for `_add_labels` that show all the different paths work as expected.  You can mock `self._get_namespace()` to return a specific namespace object, and then ensure the namespace gets patched as expected

src/charm.py
```diff
@@ -6,38 +6,231 @@
...
+            ) != f"{self.app.name}-{self.model.name}":
+                logger.error(
+                    f"Cannot add labels: Namespace '{self.model.name}' is already configured with Istio labels managed by another entity."
+                )
+                return
+
+            labels_to_add = {
+                "istio.io/use-waypoint": f"{self.app.name}-{self.model.name}-waypoint",
+                "istio.io/dataplane-mode": "ambient",
+                "istio.io.waypoint-managed-by": f"{self.app.name}-{self.model.name}",
```
ca-scribner, line 204: Can we move this label name to a constant?  that way its not a literal in multiple places

ca-scribner, line 204: label names follow the [schema](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set): `[prefix/]name`, where the prefix denotes ownership.  So in this case, what something like:
* `charms.canonical.com/istio.io.waypoint.managed-by`
* `istio-beacon-k8s-charm.canonical.com/istio.io.waypoint.managed.by`
or something like that.  I'm flexible on what, but we should have a prefix that points at canonical or something about the charm

src/charm.py
```diff
@@ -6,38 +6,231 @@
...
+            }
+
+            namespace.metadata.labels.update(labels_to_add)
+            self._patch_namespace(namespace)
+
+    def _remove_labels(self):
```
ca-scribner, line 210: Same comment as above for _add_labels.  I think this might break, but either way lets add tests to show what it does.

src/charm.py
```diff
@@ -6,38 +6,231 @@
...
+        )
+
+    def _sync_waypoint_resources(self):
+        resources_list = []
+        krm = self._get_waypoint_resource_manager()
+        resource_to_append = self._construct_waypoint()
+        resources_list.append(resource_to_append)
+        krm.reconcile(resources_list)
+
+        if self.config["model-on-mesh"]:
```
ca-scribner, line 162: Can we do a unit or scenario test that confirms, when model-on-mesh is set T/F, we do/do-not try to add/remove the labels on config changed.  I think it can be pretty high level where we mock the lightkube client, run the event, then assert on whether `lightkube.patch(namespace, ...)` was called

tests/integration/test_charm.py
```diff
@@ -14,17 +17,49 @@
...
+    await istio_beacon.model.wait_for_idle(
+        [APP_NAME], status="active", timeout=1000, raise_on_error=False
+    )
+    await validate_labels(istio_beacon, APP_NAME, should_be_present=True)
+
+    await istio_beacon.model.applications[APP_NAME].set_config({"model-on-mesh": "false"})
+    await istio_beacon.model.wait_for_idle(
+        [APP_NAME], status="active", timeout=1000, raise_on_error=False
     )
+    await validate_labels(istio_beacon, APP_NAME, should_be_present=False)
```
ca-scribner, line 65: This is a correct test.  I'm torn on whether it should be moved to a unit test though.  As an integration test it takes longer.

I lean toward using a unit test here.  We can model the behaviour accurately in a unit test, and then it runs in a second without Juju deployed at all.  wdyt?

ca-scribner (commented):
some suggestions but overall looks close!

src/charm.py
```diff
@@ -6,38 +6,231 @@
...
+    def _sync_all_resources(self):
+        if not self.unit.is_leader():
+            self.unit.status = BlockedStatus("Waypoint can only be provided on the leader unit.")
+            return
+        self.unit.status = MaintenanceStatus("Validating waypoint readiness")
+        self._sync_waypoint_resources()
+        if not self._is_ready():
+            self.unit.status = BlockedStatus(
+                "Waypoint's k8s deployment not ready, is istio properly installed?"
+            )
```
dstathis, line 126: This should not be BlockedStatus. ΒlockedStatus means "human intervention required". This should be error status (set by raising an exception). The charm has no idea what went wrong and whether a human should be fixing it. An additional benefit of this would be that juju will retry the hook automatically.

src/charm.py
```diff
@@ -6,38 +6,231 @@
...
+                ):
+                    return True
+            except ApiError as e:
+                logger.error(f"Error checking waypoint deployment status: {e}")
+
+            time.sleep(check_interval)
+
+        return False
+
+    def _is_ready(self) -> bool:
```
dstathis, line 111: Since we are likely to deploy different kinds of deployments in the future, I think a better name would be `_is_waypoint_ready`.

charmcraft.yaml
```diff
@@ -43,3 +43,22 @@ parts:
...
+config:
+  options:
+    ready-timeout:
+      type: int
+      default: 100
+      description: >
+        The maximum time (in seconds) to wait for the waypoint deployment to be 
+        ready. This applies specifically to the deployment created for the Istio 
+        waypoint controller. If the deployment does not become ready within this time, 
+        charm will go into blocked state.
```
dstathis, line 58: See my other comment about `BlockedStatus`.

src/charm.py
dstathis: Yeah I think `MaintenanceStatus` is the way to go here.

tests/integration/test_charm.py
IbraAoad: I added the unit tests needed, yet I'd also keep the itests to test k8s validation webhooks on labels and its behavior with the whole `peer connection reset` thing, since this is practically a new area for us I want to be 100% sure that it will always go into active on hook retry

src/charm.py
IbraAoad: Not sure I get you guys, doesn't the `MaintenanceStatus("Validating waypoint readiness")` serve this purpose?, since the status here will stick until it actually goes into blocked state if not ready or Active if ready?

src/charm.py
IbraAoad: Tbh I'd argue that `BlockedStatus` is the right way to go here since it actually needs a human intervention, this mainly was added here to tackle the scenario in which a charm author deploys this charm before deploying `istio-k8s` which deploys the CRDs so an actual "human" needs to go in and install it.

also the hook retrying in this context wouldn't be beneficial as we already spent a decent amount of time waiting on the waypoint to be ready `ready-timeout` which is also configurable and in case this isn't enough time, a "human" should also intervene here and increase the config time.

charmcraft.yaml
IbraAoad: replied on the other comment

src/charm.py
IbraAoad: oh actually on second thought, I think your argument is correct and my train of thoughts might have tricked me here.

even if `istio-k8s` isn't installed and a human intervention is required to install it, the beacon won't have any knowledge of this as they won't be related and the charm will stay in blocked state.
will switch to an error status, please ignore my previous comment :smile_cat:

.github/workflows/pull-request.yaml
```diff
@@ -5,8 +5,9 @@ on:
     branches:
       - main
 
+# TODO: Remove when #8 closes
```
dstathis, line 8: This comment could be a bit more clear. Remove what?

dstathis (approved):

ca-scribner (approved):
some suggestions for the tests, but lgtm

PR #12, opened 2024-09-20
feat: generate `AuthorizationPolicies` based on service-mesh relation data

ca-scribner:
## Issue
Adds the feature of creating `AuthorizationPolicies` based on the data provided from applications related on the `service-mesh` relation.  

## Testing instructions
WIP: Integration tests still WIP.  See the integration tests which use a tester charm that can be either side of the service mesh relation.  

## Drive-bys

Some changes are made to the service_mesh.py library based on issues hit during implementing this PR.


src/charm.py
```diff
@@ -119,12 +156,75 @@ def _sync_all_resources(self):
...
+        for i, policy in enumerate(mesh_info):
+            authorization_policies[i] = RESOURCE_TYPES["AuthorizationPolicy"](
+                metadata=ObjectMeta(
+                    # TODO: Each related app can define multiple policies that apply to the same source/target pair -
+                    #  how do we name the policies below to account for that?  For now, we suffix the index of the
+                    #  policy in the data bag, but that isn't great because it means a policy is added to the start of
+                    #  the list (or even just reorder the existing list) then all other policies are renamed causing
+                    #  them to be deleted and recreated.
+                    #  Naming these with just source_app_name-target_app_name isn't enough, is it?  I think we need to
+                    #  support multiple policies per source/target pair?
```
ca-scribner, line 182: This needs more work, whether we do it now or park it for future.  I'm open to suggestions.  Haven't spent a ton of thought on it yet, maybe there's a simple solution

src/charm.py
```diff
@@ -119,12 +156,75 @@ def _sync_all_resources(self):
...
+                                    )
+                                )
+                                for endpoint in policy.endpoints
+                            ],
+                        )
+                    ],
+                    # by_alias=True because the model includes an alias for the `from` field
+                    # exclude_unset=True because unset fields will be treated as their default values in Kubernetes
+                    # exclude_none=True because null values in this data always mean the Kubernetes default
+                ).model_dump(by_alias=True, exclude_unset=True, exclude_none=True),
```
ca-scribner, line 224: I've made most of the fields in our data model optional, which then means we need to be careful how we export them here.  I'm not sure now if making these fields optional was required - curious what others think

lib/charms/istio_beacon_k8s/v0/service_mesh.py
```diff
@@ -140,23 +190,78 @@ def update_service_mesh(self):
...
+        for relation in self._charm.model.relations[self._relation_name]:
+            relation.data[self._charm.app]["labels"] = rel_data
+
+    def mesh_info(self) -> List[MeshPolicy]:
+        """Return the relation data used to define authorization policies on the mesh."""
+        mesh_info = []
+        for relation in self._charm.model.relations[self._relation_name]:
+            policies_data = json.loads(relation.data[relation.app]["policies"])
+            policies = [MeshPolicy.model_validate(policy) for policy in policies_data]
+            mesh_info.extend(policies)
```
ca-scribner, line 256: This edit will be moved into #13

lib/charms/istio_beacon_k8s/v0/service_mesh.py
```diff
@@ -140,23 +190,78 @@ def update_service_mesh(self):
...
+
+        # TODO: Where should we put this default resolution?  Could be in Requirer side, here on Provider, or in charm
```
ca-scribner, line 258: Per discussion with @dstathis, we'll move this logic into the charm itself rather than the lib.  The reason being that, in theory, an istio charm might have a different idea of what a default value should be than say a cilium charm, etc.

tests/scenario/test_authorization_policies.py
```diff
@@ -0,0 +1,87 @@
...
+                            )
+                        ],
+                    ).model_dump(),
+                ]
+            )
+        },
+    )
+
+
+def test_get_authorization_policies_from_related_apps(
```
ca-scribner, line 54: This really is a `mesh_info` test, and should be moved to #13

tests/integration/testers/service-mesh-tester/src/charm.py
```diff
@@ -0,0 +1,72 @@
+#!/usr/bin/env python3
+# Copyright 2022 Canonical Ltd.
+# See LICENSE file for licensing details.
+import logging
+
+import ops
+from charms.istio_beacon_k8s.v0.service_mesh import Endpoint, Policy, ServiceMeshRequirer
```
IbraAoad, line 7: not sure if there might be smth wrong in packing, but I get the below when using the tester charm

```bash
unit-sender-0: 12:19:51 WARNING unit.sender/0.install Traceback (most recent call last):
unit-sender-0: 12:19:51 WARNING unit.sender/0.install   File "./src/charm.py", line 7, in <module>
unit-sender-0: 12:19:51 WARNING unit.sender/0.install     from charms.istio_beacon_k8s.v0.service_mesh import Endpoint, Policy, ServiceMeshRequirer
unit-sender-0: 12:19:51 WARNING unit.sender/0.install ModuleNotFoundError: No module named 'charms.istio_beacon_k8s'
unit-sender-0: 12:19:52 ERROR juju.worker.uniter.operation hook "install" (via hook dispatching script: dispatch) failed: exit status 1
```

src/charm.py
dstathis: I think we should just open an issue for it for now as it will still work as is. It may just have some small interruptions when relations are changed. Perhaps we could append some kind of short hash of the policy contents? It depends how human readable we want it to be.

tests/integration/testers/service-mesh-tester/src/charm.py
dstathis: I assume this is because the integration tests probably copy the lib from the main charm repo rather than committing a second copy of this, so for building manually, you likely need to copy the lib. I would say we should put a symlink there but I think charmcraft can't handle symlinks.

dstathis (approved):

src/charm.py
```diff
@@ -150,6 +250,14 @@ def _construct_waypoint(self):
             spec=gateway.spec.model_dump(),
         )
 
+    def _sync_authorization_policies(self):
+        """Sync authorization policies."""
+        krm = self._get_authorization_policy_resource_manager()
+        authorization_policies = self._build_authorization_policies(self._mesh.mesh_info())
+        logger.debug("Reconciling state of AuthorizationPolicies to:")
+        logger.debug(authorization_policies)
```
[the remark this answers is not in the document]
dstathis: I like it.

tests/scenario/test_authorization_policies.py
dstathis: Since that is merged already, we can resolve this conversation.

tests/integration/testers/service-mesh-tester/src/charm.py
ca-scribner: yeah I think @dstathis has it right.  If you do the actual integration tests, it'll copy over the lib.  This way we don't use a stale lib.  But then if you manually pack, you need to copy it yourself

I'm open to suggestions on how to do it differently, but I couldn't think of a better way

tests/integration/test_charm.py
```diff
@@ -35,31 +36,90 @@ class CharmDeploymentConfiguration:
...
+    )
+    await ops_test.model.deploy(
+        sender_receiver_charm, application_name="sender1", resources=resources
+    )
+    await ops_test.model.add_relation("receiver1:service-mesh", APP_NAME)
+    await ops_test.model.add_relation("receiver1:inbound", "sender1:outbound")
+
+    await ops_test.model.wait_for_idle([APP_NAME])
+
+    await ops_test.model.wait_for_idle(["receiver1", "sender1", "sender2"])
```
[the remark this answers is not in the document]
ca-scribner: Good question.  Poor `sender2` got refactored out by mistake :D  Added it back

src/charm.py
```diff
@@ -121,12 +155,84 @@ def _sync_all_resources(self):
...
+        authorization_policies = [None] * len(mesh_info)
+        for i, policy in enumerate(mesh_info):
+            target_service = policy.target_service
+            if target_service is None:
+                target_service = policy.target_app_name
+                logger.info(
+                    f"Got policy for application '{policy.target_app_name}' that has no target_service. Defaulting to"
+                    f" application name '{target_service}'."
+                )
+            target_service = policy.target_service if policy.target_service else policy.target_app_name
```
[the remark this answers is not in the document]
ca-scribner: good improvement!

Issue #4, opened 2024-07-17
Better example in service mesh lib.


dstathis:
### Enhancement Proposal

The example in the service_mesh lib docs is not very in depth. Lets make it more detailed.

ca-scribner:
todo from #13:
* [ ] https://github.com/canonical/istio-beacon-k8s-operator/pull/13#discussion_r1773582345
* [ ] https://github.com/canonical/istio-beacon-k8s-operator/pull/13#discussion_r1773585120
* [ ] https://github.com/canonical/istio-beacon-k8s-operator/pull/13/files#r1778980694

dstathis:
This was completed by #82.

PR #13, opened 2024-09-23
Lib work

dstathis:
## Issue
Update the library to provide the current desired interface and functionality.

## Solution
Fully implement cross model support and simplify the service_mesh interface.

## Testing Instructions
 * Deploy beacon
 * Deploy  https://github.com/dstathis/mesh-tester
 * Repeat the previous step as much as you want
 * Create various relations between the tester charms
 * Relate one of the tester charm to beacon
 * Check that `jhack show-relation beacon tester` has appropriate data


lib/charms/istio_beacon_k8s/v0/service_mesh.py
```diff
@@ -6,6 +6,8 @@
 The service mesh library is used to facilitate adding your charmed application to a service mesh.
 The library leverages the `service_mesh` and `cross_model_mesh` interfaces.
 
+##Consumer
+
```
ca-scribner, line 10: I think "consumer" is more understandable than "requirer", but we should at least say this implements the requirer side of the relation.  Yes, they can deduce this must be requirer if the read down to see a "provider" later, but it should be more explicit.  

We should also include instructions on actually adding this relation to the charmcraft.yaml (like [here](https://github.com/canonical/grafana-agent-k8s-operator/blob/5b35434cffb3af185f68fb2fbacddbb1cdea114a/lib/charms/grafana_k8s/v0/grafana_dashboard.py#L24)

lib/charms/istio_beacon_k8s/v0/service_mesh.py
```diff
@@ -34,16 +36,43 @@
 except ops.TooManyRelatedAppsError as e:
     self.unit.status = BlockedStatus(e)
 ```
+
+You will then receive the labels which you need to add your product to the mesh:
+```
+def _on_mesh_relation_changed(self, event):
+    self._apply_labels to pods(self._mesh.labels())
```
ca-scribner, line 43:
```suggestion
    self._apply_labels_to_pods(self._mesh.labels())
```

tests/scenario/test_mesh_consumer.py
```diff
@@ -0,0 +1,148 @@
...
+            "endpoints": [{"hosts": [], "ports": [80], "methods": [], "paths": []}],
+        }
+    ],
+)
+
+POLICY_DATA_PARAMS = [SIMPLE, CMR, REVERSE, REVERSE_CMR]
+
+
+@pytest.mark.parametrize("policies,expected_data", POLICY_DATA_PARAMS)
+def test_relation_data_policies(policies, expected_data):
```
ca-scribner, line 99: A lot goes on in this test.  Can you add a docstring or comment explaining what is being constructed here and how it should behave?

tests/scenario/test_mesh_consumer.py
```diff
@@ -0,0 +1,148 @@
...
+            "source_namespace": "remote_model",
+            "target_app_name": "consumer-charm",
+            "target_namespace": "my_model",
+            "target_service": None,
+            "endpoints": [{"hosts": [], "ports": [80], "methods": [], "paths": []}],
+        }
+    ],
+)
+
+REVERSE = (
```
ca-scribner, line 67: Can you add descriptions of what these setups should demonstrate in the tests?  SIMPLE and CMR are fairly clear, but I'm not sure about REVERSE*

tests/scenario/test_mesh_consumer.py
```diff
@@ -0,0 +1,148 @@
...
+from charms.istio_beacon_k8s.v0.service_mesh import Endpoint, Policy, ServiceMeshConsumer
+from ops import CharmBase
+
+
+def consumer_context(policies: typing.List[Policy]):
+    meta = {
+        "name": "consumer-charm",
+        "requires": {
+            "service-mesh": {"interface": "service_mesh"},
+            "require-cmr-mesh": {"interface": "cross_model_mesh"},
```
ca-scribner, line 18: What does this relation do?  I dont see it used anywhere

tests/scenario/test_mesh_consumer.py
```diff
@@ -0,0 +1,148 @@
...
+        def __init__(self, framework):
+            super().__init__(framework)
+            self.mesh = ServiceMeshConsumer(charm=self, policies=policies)
+
+    return scenario.Context(ConsumerCharm, meta)
+
+
+ENDPOINT_A = Endpoint(hosts=[], ports=[80], methods=[], paths=[])
+
+SIMPLE = (
```
ca-scribner, line 39: Can we include in some of these cases multiple ports, some methods, paths, and a case that has multiple policies?

tests/scenario/test_mesh_consumer.py
```diff
@@ -0,0 +1,148 @@
...
+from ops import CharmBase
+
+
+def consumer_context(policies: typing.List[Policy]):
+    meta = {
+        "name": "consumer-charm",
+        "requires": {
+            "service-mesh": {"interface": "service_mesh"},
+            "require-cmr-mesh": {"interface": "cross_model_mesh"},
+            "rela": {"interface": "foo"},
```
ca-scribner, line 19: There's a lot of string literals here (`rela`, `remote_a`, `my_model`, etc) that are written in several places but must be consistent.  Can we change these to some constants that get used everywhere?

ca-scribner (commented):
From what I follow things look good, but there's areas where I'm not sure what the intent is.  Can you add more description to the PR and maybe the implementation describing how things should be used?  In particular, I'm uncertain about the CMR parts

lib/charms/istio_beacon_k8s/v0/service_mesh.py
```diff
@@ -78,37 +107,65 @@ class Endpoint(pydantic.BaseModel):
...
+        self._cmr_relations = self._charm.model.relations[cross_model_mesh_provides_name]
         self._policies = policies or []
         self.framework.observe(
             self._charm.on[mesh_relation_name].relation_created, self._relations_changed
         )
+        self.framework.observe(
+            self._charm.on[cross_model_mesh_requires_name].relation_created, self._send_cmr_data
+        )
         self.framework.observe(self._charm.on.upgrade_charm, self._relations_changed)
         for policy in self._policies:
```
ca-scribner, line 170: If I have two policies for the same relation, does this double-subscribe the event (eg: we'll do a `relation_*` twice for each event)?  I'm not sure if ops deduplicates this

lib/charms/istio_beacon_k8s/v0/service_mesh.py
```diff
@@ -130,33 +194,96 @@ def update_service_mesh(self):
...
+            relation.data[self._charm.app]["labels"] = rel_data
+
+    def mesh_info(self):
+        """Return the relation data used to define authorization policies on the mesh."""
+        mesh_info = []
+        for relation in self._charm.model.relations[self._relation_name]:
+            rel_data = json.loads(relation.data[relation.app]["mesh_data"])
+            rel_data["policies"] = [Policy.parse_obj(p) for p in rel_data["policies"]]
+            mesh_info.append(rel_data)
+        return mesh_info
```
ca-scribner, line 289: While doing #12 I found the output from here wasn't quite what we probably want.  I modified this to be:

```suggestion
    def mesh_info(self) -> List[MeshPolicy]:
        """Return the relation data used to define authorization policies on the mesh."""
        mesh_info = []
        for relation in self._charm.model.relations[self._relation_name]:
            policies_data = json.loads(relation.data[relation.app]["policies"])
            policies = [MeshPolicy.model_validate(policy) for policy in policies_data]
            mesh_info.extend(policies)

        # TODO: Where should we put this default resolution?  Could be in Requirer side, here on Provider, or in charm
        #       that uses provider
        for policy in mesh_info:
            if policy.target_service is None:
                logger.info(
                    f"Got policy for application '{policy.target_app_name}' that has no target_service. Defaulting to"
                    f" application name '{policy.target_app_name}'."
                )
                policy.target_service = policy.target_app_name
        return mesh_info
```

We can discuss it here or in #12.  It might make more sense to discuss it in #12 though as this function has test coverage there

lib/charms/istio_beacon_k8s/v0/service_mesh.py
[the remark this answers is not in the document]
dstathis: Most charms won't need to do this. It is only for "k8s native" type charms. So I don't think we really need to explain this. Also, these docs will pretty much be rewritten as we have a separate task for it.

lib/charms/istio_beacon_k8s/v0/service_mesh.py
dstathis: Will we call observe multiple times? Yes. Will that cause ops to trigger multiple events? Not sure but I am guessing yes. Do we care? Not sure but I am guessing no.

lib/charms/istio_beacon_k8s/v0/service_mesh.py
```diff
@@ -130,33 +194,96 @@ def update_service_mesh(self):
...
                     )
-        mesh_rel_data = {
-            "app_name": self._charm.app.name,
-            "model": self._my_namespace(),
-            "policies": policies,
-        }
-        self._relation.data[self._charm.app]["mesh_data"] = json.dumps(mesh_rel_data)
+        self._relation.data[self._charm.app]["policies"] = json.dumps(mesh_policies)
+
+    def _check_cmr(self, mesh_relation: Relation) -> Optional[Relation]:
```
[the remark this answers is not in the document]
dstathis: Done.

lib/charms/istio_beacon_k8s/v0/service_mesh.py
```diff
@@ -130,33 +194,96 @@ def update_service_mesh(self):
...
         """Return the namespace of the running charm."""
         # This method currently assumes the namespace is the same as the model name. We
         # should consider if there is a better way to do this.
         return self._charm.model.name
+
+    def labels(self):
+        """Labels required for a pod to join the mesh."""
+        if self._relation is None:
+            return {}
+        return json.loads(self._relation.data[self._relation.app]["labels"])
```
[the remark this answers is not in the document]
dstathis: Great catch! Fixed.

src/charm.py
```diff
@@ -237,6 +239,10 @@ def _remove_labels(self):
             namespace.metadata.labels.update(labels_to_remove)
             self._patch_namespace(namespace)
 
+    def mesh_labels(self):
+        """Labels required for a workload to join the mesh."""
+        return {"foo": "bar"}
```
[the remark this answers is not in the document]
dstathis: Planned as part of a later PR.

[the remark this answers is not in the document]
ca-scribner: Lets add a todo here to indicate that

lib/charms/istio_beacon_k8s/v0/service_mesh.py
ca-scribner: Yeah if I had to pick something, I'd guess this has no negative consequence.  But its an unfounded guess.

I'd feel more comfortable if we just deduplicate it first.  Something like:
```python
relations = set([policy.relation for policy in self._policies])
for relation in relations:
  ...
```

lib/charms/istio_beacon_k8s/v0/service_mesh.py
```diff
@@ -34,16 +36,43 @@
...
+...
+self._mesh = ServiceMeshProvider(
+    charm = self,
+    labels = self._get_mesh_labels(),
+    mesh_relation_name = "service-mesh",
+)
+```
+
+The labels argument should be a dict containing the Kubernetes labels which the client charm needs
+to apply to join the mesh.
```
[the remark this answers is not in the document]
dstathis: The consumer does not need to do anything with these labels for a normal charm so it's just the provider side that needs to worry about this. I'm not sure I see the confusion. Could you suggest a better wording?

lib/charms/istio_beacon_k8s/v0/service_mesh.py
```diff
@@ -78,37 +107,65 @@ class Endpoint(pydantic.BaseModel):
...
 
         Args:
             charm: The charm instantiating this object.
             mesh_relation_name: The relation name as defined in metadata.yaml or charmcraft.yaml
-                for the relation which used the service_mesh interface.
+                for the relation which uses the service_mesh interface.
+            cross_model_mesh_requires_name: The relation name as defined in metadata.yaml or
+                charmcraft.yaml for the relation which requires the cross_model_mesh interface.
+            cross_model_mesh_provides_name: The relation name as defined in metadata.yaml or
+                charmcraft.yaml for the relation which provides the cross_model_mesh interface.
```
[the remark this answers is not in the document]
dstathis: I think this is clear enough in the lib documentation now. Do you still find it confusing.

lib/charms/istio_beacon_k8s/v0/service_mesh.py
ca-scribner: agreed re "consumer" vs "requirer"

For documentation, I'm linking this comment in #4 so we don't forget it later.

lib/charms/istio_beacon_k8s/v0/service_mesh.py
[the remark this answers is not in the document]
ca-scribner: Added a note about this in #4.  we can close it here

lib/charms/istio_beacon_k8s/v0/service_mesh.py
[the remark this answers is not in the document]
ca-scribner: what about:

```suggestion
The labels argument is the dict of labels that indicate to the service mesh that a Pod should be subscribed to the mesh.
These will be sent to each related ServiceMeshConsumer to be used by the charm for per-Pod subscription.  These labels
are service-mesh dependent - for example, for an Istio ambient mesh this should be 
{"istio.io/dataplane-mode": "ambient"}.  
```

?

and then something similar in the docstring of the class itself

lib/charms/istio_beacon_k8s/v0/service_mesh.py
```diff
@@ -130,33 +218,97 @@ def update_service_mesh(self):
...
+                return cmr_rel
+        return None
 
     def _my_namespace(self):
         """Return the namespace of the running charm."""
         # This method currently assumes the namespace is the same as the model name. We
         # should consider if there is a better way to do this.
         return self._charm.model.name
+
+    def labels(self):
```
ca-scribner, line 269: Istio (and I assume others?) have labels both to opt into **or out of** the service mesh (`istio.io/dataplane-mode=ambient` vs `istio.io/dataplane-mode=none`).  The latter is important because some workloads just don't function with the service mesh involved (there's examples in kubeflow, db's, etc).  

Adding opt-out labels feels out of scope until we need it, but could we rename `labels` here and in the docs etc to something else, maybe `opt_in_labels` (I bet there's a better name, just cant think of one now)?  `Consumer.labels()` will be part of the public api, so naming it in a way that wont be confusing if there's opt-out labels feels prudent.  wdyt?

lib/charms/istio_beacon_k8s/v0/service_mesh.py
```diff
@@ -34,16 +36,43 @@
...
+)
+```
+
+The labels argument should be a dict containing the Kubernetes labels which the client charm needs
+to apply to join the mesh.
+
+You can then use the relation data to build your authorization policies:
+```
+self._build_authorization_policies(self._mesh.mesh_info())
+```
```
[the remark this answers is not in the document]
ca-scribner: I think this is still outstanding.  Authorization policies are an Istio thing and the service-mesh lib is mesh agnostic - this feels out of place

lib/charms/istio_beacon_k8s/v0/service_mesh.py
```diff
@@ -6,44 +6,97 @@
 The service mesh library is used to facilitate adding your charmed application to a service mesh.
 The library leverages the `service_mesh` and `cross_model_mesh` interfaces.
 
-To add service mesh support to your charm, instantiate a ServiceMeshConsumer object in the
+##Consumer
+
+To add service mesh support to your charm, you must add 3 relations in your charmcraft.yaml.
```
ca-scribner, line 11: atm this docstring feels really clear to someone who knows what Istio does and its terminology (mainly, authorization policy), but feels confusing to a newcomer.  What about some more intro material up front?

```suggestion
Service meshes provide a range of capabilities for routing, controlling, and monitoring traffic.  A key feature of many service meshes is the ability to restrict traffic between Pods based on L4 and L7 criteria.  For example, defining that a Pod MetricsScraper can `GET` from Pod MetricsProducer at `/metrics` on port `9090`, but SomeOtherPod cannot.  

The ServiceMeshConsumer object is used to subscribe a charm and its workloads to a related service mesh.  Since it is common for a relation between applications to indicate traffic flow (ex: if DbConsumer Requires a DbProducer), the ServiceMeshConsumer provides an optional way to automate creation of traffic rules based on app relations. 

To add service mesh support to your charm, you must add 3 relations in your charmcraft.yaml.
```

ca-scribner: We can park this for now and address it in #4 if you'd like, just making a note of it now

lib/charms/istio_beacon_k8s/v0/service_mesh.py
```diff
@@ -6,44 +6,97 @@
...
+)
+```
+
+The example above specifies two policies. The resulting behaviour would be that when related over the `metrics` relation, the service mesh will allow traffic to the `/metrics` endpoint for the remote application and when related over the `data` endpoint, the service mesh will allow traffic to the `/data` endpoint.
+
+By using the above method, you can specify exactly which endpoints can be reached over which relations.
+
+###Cross Model relations
+
+If a relation is cross model, and addition step is required to make the authorization policies work. You must also relate the two applications over the cross_model_mesh interface. When that relation is established, traffic will be allowed from the requirer to the provider.
```
ca-scribner, line 67: just a suggestion

```suggestion
If a ServiceMeshConsumer is creating a Policy for a relation that is cross model, additional information is required to construct the policy. To facilitate this, the charms can also be related over the cross_model_mesh interface. When that relation is established, traffic will be allowed from the requirer to the provider.
```

lib/charms/istio_beacon_k8s/v0/service_mesh.py
```diff
@@ -130,33 +218,97 @@ def update_service_mesh(self):
...
+        self.update_relations()
+
+    def update_relations(self):
+        """Update all relations with the labels needed to use the mesh."""
+        rel_data = json.dumps(self._labels)
+        for relation in self._charm.model.relations[self._relation_name]:
+            relation.data[self._charm.app]["labels"] = rel_data
+
+    def mesh_info(self) -> List[MeshPolicy]:
+        """Return the relation data used to define authorization policies on the mesh."""
```
ca-scribner, line 308: removing "authorization" since its mesh agnostic.  Exact wording is just a suggestion

```suggestion
        """Return the relation data that defines Policies requested by the related applications."""
```

lib/charms/istio_beacon_k8s/v0/service_mesh.py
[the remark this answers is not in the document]
ca-scribner: yeah better now.  ty!

lib/charms/istio_beacon_k8s/v0/service_mesh.py
```diff
@@ -130,33 +218,97 @@ def update_service_mesh(self):
...
-        self._relation.data[self._charm.app]["mesh_data"] = json.dumps(mesh_rel_data)
+        self._relation.data[self._charm.app]["policies"] = json.dumps(mesh_policies)
+
+    def _check_cmr(self, mesh_relation: Relation) -> Optional[Relation]:
+        """Check if the given relation is a cmr. If so return the associated cross_model_mesh relation."""
+        for cmr_rel in self._cmr_relations:
+            if cmr_rel.app.name == mesh_relation.app.name:
+                if "cmr_data" not in cmr_rel.data[cmr_rel.app]:
+                    # Data has not yet been provided
+                    continue
```
ca-scribner, line 259: Is `continue` needed here?  I don't think there's a case where we should have a match, find no data, and then loop to find a second match, is there?

lib/charms/istio_beacon_k8s/v0/service_mesh.py
```diff
@@ -130,33 +218,97 @@ def update_service_mesh(self):
...
-            "model": self._my_namespace(),
-            "policies": policies,
-        }
-        self._relation.data[self._charm.app]["mesh_data"] = json.dumps(mesh_rel_data)
+        self._relation.data[self._charm.app]["policies"] = json.dumps(mesh_policies)
+
+    def _check_cmr(self, mesh_relation: Relation) -> Optional[Relation]:
+        """Check if the given relation is a cmr. If so return the associated cross_model_mesh relation."""
+        for cmr_rel in self._cmr_relations:
+            if cmr_rel.app.name == mesh_relation.app.name:
```
ca-scribner, line 256: What happens if I have
* model1
  * TargetApp
* model2
  * SourceApp
* model3
  * SourceApp

where both SourceApps relate to TargetApp?  I think this comparison wont distinguish between these.  

We should add a test for this situation - its an easy edge case to miss.  Then it'll be clear if we're affected by it.

lib/charms/istio_beacon_k8s/v0/service_mesh.py
```diff
@@ -130,33 +218,97 @@ def update_service_mesh(self):
...
-        cmr_matcher = re.compile(r"remote\-[a-f0-9]+")
+        mesh_policies = []
         for policy in self._policies:
             for relation in self._charm.model.relations[policy.relation]:
-                if cmr_matcher.fullmatch(relation.app.name):
-                    logger.debug(
-                        f"Cross model relation found: {relation.name}. Currently not implemented. Skipping."
+                if cmr_relation := self._check_cmr(relation):
+                    logger.debug(f"Found cross model relation: {relation.name}. Creating policy.")
+                    cmr_data = CMRData.parse_obj(
```
ca-scribner, line 226:
```suggestion
                    cmr_data = CMRData.model_validate(
```

lib/charms/istio_beacon_k8s/v0/service_mesh.py
```diff
@@ -130,33 +218,97 @@ def update_service_mesh(self):
...
                     )
-        mesh_rel_data = {
-            "app_name": self._charm.app.name,
-            "model": self._my_namespace(),
-            "policies": policies,
-        }
-        self._relation.data[self._charm.app]["mesh_data"] = json.dumps(mesh_rel_data)
+        self._relation.data[self._charm.app]["policies"] = json.dumps(mesh_policies)
+
+    def _check_cmr(self, mesh_relation: Relation) -> Optional[Relation]:
```
ca-scribner, line 253: what about renaming this to `_get_cmr` or `_get_cmr_for_relation`?  "check_cmr" sounds like we're validating the CMR, when really this is more a lookup of the associated cmr

lib/charms/istio_beacon_k8s/v0/service_mesh.py
ca-scribner: This feels like something to `raise` on and catch in the calling function, at least to write a log.  At minimum the connection wont be functioning properly if we're hitting this case

tests/integration/test_charm.py
ca-scribner, line 1: can we add a cross-model integration test that does a simple demonstration that the right things happen for CMR?

ca-scribner (commented):
getting close.  Added a few comments and there's a few outstanding

lib/charms/istio_beacon_k8s/v0/service_mesh.py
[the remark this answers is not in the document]
dstathis: Okay I took the example out and just generally mentioned the method.

src/charm.py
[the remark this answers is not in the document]
dstathis: Done.

tests/scenario/test_mesh_consumer.py
dstathis: Done.

tests/scenario/test_mesh_consumer.py
dstathis: I instead changed to names to be more descriptive. Let me know if you think it's good now.

tests/scenario/test_mesh_consumer.py
dstathis: This is the relation between the consumer and beacon. It is called "mesh_relation" in the test and is used in the assert statement.

lib/charms/istio_beacon_k8s/v0/service_mesh.py
dstathis: Agreed and fixed.

lib/charms/istio_beacon_k8s/v0/service_mesh.py
dstathis: It is both a check and a get so neither name is really "correct".

lib/charms/istio_beacon_k8s/v0/service_mesh.py
dstathis: We are not affected by it. The app name is always going to be unique *in model1*.

lib/charms/istio_beacon_k8s/v0/service_mesh.py
dstathis: You are correct and it has been fixed.

lib/charms/istio_beacon_k8s/v0/service_mesh.py
dstathis: The seems fine. We can always have labels and opt-out-labels later.

tests/scenario/test_mesh_consumer.py
dstathis: I see now that you mean the "require-cmr-mesh" relation. You are correct that it is not used but I want to leave it in the metadata as an actual charm would have it and it may be used in future tests.

tests/scenario/test_mesh_consumer.py
dstathis: Tests added.

IbraAoad (approved):

tests/integration/test_charm.py
dstathis: We need a tester charm before we can do this. I will open an issue.

PR #18, opened 2024-10-11
Join mesh

dstathis:
## Issue
We need the service_mesh library to automatically join the charm to the mesh.

## Solution
We apply the labels to the underlying pods.

## Testing Instructions
* Deploy istio-k8s
* Deploy beacon
* Deploy [mesh-tester](https://github.com/dstathis/mesh-tester)
* Relate beacon to mesh-tester
* Check for the Istio labels

### DIFF
--- a/lib/charms/istio_beacon_k8s/v0/service_mesh.py
+++ b/lib/charms/istio_beacon_k8s/v0/service_mesh.py
@@ -337,14 +337,7 @@ class ServiceMeshConsumer(Object):
                     labels[label] = None
         if stateful_set.spec:
             stateful_set.spec.template.metadata.labels.update(labels)
-        config_map.data = {
-            "labels": json.dumps(
-                {key: value for (key, value) in labels.items() if value is not None}
-            )
-        }
-        # TODO: We should look into whether it is possible to do the next two commands as one
-        # transaction. As it is now, if the code crashes between the two commands, we could end up
-        # leaving dangling labels on the StatefulSet.
+        config_map.data = {"labels": json.dumps(labels)}
         client.patch(res=ConfigMap, name=self._label_configmap_name, obj=config_map)
         client.patch(res=StatefulSet, name=self._charm.app.name, obj=stateful_set)
 
--- a/lib/charms/istio_beacon_k8s/v0/service_mesh.py
+++ b/lib/charms/istio_beacon_k8s/v0/service_mesh.py
@@ -70,6 +70,10 @@ By using the above method, you can specify exactly which endpoints can be reache
 
 If a ServiceMeshConsumer is creating a Policy for a relation that is cross model, additional information is required to construct the policy. To facilitate this, the charms can also be related over the cross_model_mesh interface. When that relation is established, traffic will be allowed from the requirer to the provider.
 
+###Joining the Mesh
+
+For most charms, simply instantiating ServiceMeshConsumer should automatically configure the charm to be on the mesh. If your charm is the one of the old "podspec" style charms or your charm deploys custom k8s resources there is an addition step. You must apply the labels returned by `ServiceMeshConsumer.labels()` to your pods.
+
 ##Provider
 
 To provide a service mesh, instantiate the ServiceMeshProvider object in the __init__ method
@@ -104,13 +108,18 @@ import logging
 from typing import Dict, List, Optional
 
 import pydantic
+from lightkube.core.client import Client
+from lightkube.core.exceptions import ApiError as LightkubeApiError
+from lightkube.models.meta_v1 import ObjectMeta
+from lightkube.resources.apps_v1 import StatefulSet
+from lightkube.resources.core_v1 import ConfigMap
 from ops import CharmBase, Object, Relation
 
 LIBID = "3f40cb7e3569454a92ac2541c5ca0a0c"  # Never change this
 LIBAPI = 0
 LIBPATCH = 1
 
-PYDEPS = ["pydantic"]
+PYDEPS = ["lightkube", "pydantic"]
 
 logger = logging.getLogger(__name__)
 
@@ -174,6 +183,7 @@ class ServiceMeshConsumer(Object):
         cross_model_mesh_requires_name: str = "require-cmr-mesh",
         cross_model_mesh_provides_name: str = "provide-cmr-mesh",
         policies: Optional[List[Policy]] = None,
+        auto_join: bool = True,
     ):
         """Class used for joining a service mesh.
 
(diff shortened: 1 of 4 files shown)


lib/charms/istio_beacon_k8s/v0/service_mesh.py
```diff
@@ -287,12 +306,64 @@ def _my_namespace(self):
...
+        self._delete_label_configmap()
+
+    def _update_labels(self, _event):
+        self._set_labels(self.labels())
+
+    def _set_labels(self, labels: dict) -> None:
+        client = Client(namespace=self._charm.model.name, field_manager=self._charm.app.name)
+        stateful_set = client.get(res=StatefulSet, name=self._charm.app.name)
+        try:
+            config_map = client.get(ConfigMap, self._label_configmap_name)
```
IbraAoad, line 326: I thought the configmap work would happen on the provider side, one centralized config that has key values of each charm relating to the beacon with the labels it should apply and then the consumer would apply that label on it's own statefulset if a key in configmap == app_name

Not sure if you think this would be a good approach, but in the current approach if a charm author force deleted his charm we might have tailing configmaps everywhere.

lib/charms/istio_beacon_k8s/v0/service_mesh.py
```diff
@@ -287,12 +306,64 @@ def _my_namespace(self):
...
+    def _on_mesh_broken(self, _event):
+        self._set_labels({})
+        self._delete_label_configmap()
+
+    def _update_labels(self, _event):
+        self._set_labels(self.labels())
+
+    def _set_labels(self, labels: dict) -> None:
+        client = Client(namespace=self._charm.model.name, field_manager=self._charm.app.name)
+        stateful_set = client.get(res=StatefulSet, name=self._charm.app.name)
```
IbraAoad, line 324: We might need to catch ApiErrors here as well

lib/charms/istio_beacon_k8s/v0/service_mesh.py
```diff
@@ -287,12 +306,64 @@ def _my_namespace(self):
...
+                    # The label was previously set. Setting it to None will delete it.
+                    labels[label] = None
+        if stateful_set.spec:
+            stateful_set.spec.template.metadata.labels.update(labels)
+        config_map.data = {
+            "labels": json.dumps(
+                {key: value for (key, value) in labels.items() if value is not None}
+            )
+        }
+        # TODO: We should look into whether it is possible to do the next two commands as one
```
IbraAoad, line 345: one way to do this would be
1 - save the statefulset state in a var, you already do in 
```python
stateful_set = client.get(res=StatefulSet, name=self._charm.app.name)
```
2 - put the configmap patch in a try block, if it fails we can raise and exit
3 - put the sts patch in another try block, if it fails we can override the object with the original state via lk.replace()
```python
client.replace(res=StatefulSet, name=self._charm.app.name, obj=stateful_set)
```

lib/charms/istio_beacon_k8s/v0/service_mesh.py
```diff
@@ -287,12 +306,64 @@ def _my_namespace(self):
...
+    def _update_labels(self, _event):
+        self._set_labels(self.labels())
+
+    def _set_labels(self, labels: dict) -> None:
+        client = Client(namespace=self._charm.model.name, field_manager=self._charm.app.name)
+        stateful_set = client.get(res=StatefulSet, name=self._charm.app.name)
+        try:
+            config_map = client.get(ConfigMap, self._label_configmap_name)
+        except LightkubeApiError as e:
+            if "not found" in str(e):
```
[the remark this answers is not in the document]
dstathis: Is ApiError an instance of HTTPStatusError?

lib/charms/istio_beacon_k8s/v0/service_mesh.py
dstathis: And what action would we take after catching these errors? I would rather just let them surface and have juju retry the hook automatically.

lib/charms/istio_beacon_k8s/v0/service_mesh.py
dstathis: With the centralized approach, force removing the beacon would actually still leave all the related charms on the mesh. I *think* I would prefer the chance of a dangling config map (which does nothing) to dangling labels (which do something).

dstathis: Also if the client app has custom k8s resources, it needs to be the one to handle labels. I would prefer not to have beacon taking care of it for some charms and the charm handling it for others.

lib/charms/istio_beacon_k8s/v0/service_mesh.py
dstathis: And if the replace fails?

lib/charms/istio_beacon_k8s/v0/service_mesh.py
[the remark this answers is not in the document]
IbraAoad: Yeah you probably can do smth like this as well

    except ApiError as e:
        if e.response.status_code == 404:

lib/charms/istio_beacon_k8s/v0/service_mesh.py
IbraAoad: tbh the possibilities of replace failing is very unlikely since we are sending the original state back which was already an accepted state by k8s, the only case I see it failing as this point is due to a connection error which is still to me sounds very unlikely with all the previous try blocks

lib/charms/istio_beacon_k8s/v0/service_mesh.py
IbraAoad: So I'm not proposing that everything happens on the provider side, I'm suggesting to only keep the config-map centralized and let every consumer label it's own sts, if beacon is forced removed then the consumer hooks will be run to remove each sts labels

lib/charms/istio_beacon_k8s/v0/service_mesh.py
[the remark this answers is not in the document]
dstathis: It works. Good call.

lib/charms/istio_beacon_k8s/v0/service_mesh.py
dstathis: This seems really messy to me. So you are saying beacon creates the ConfigMap and the charms modify it? Seems we would need to implement some kind of locking of the config map so charms don't write over each others changes.

lib/charms/istio_beacon_k8s/v0/service_mesh.py
dstathis: Don't these commands only really fail from a connection error anyway?

lib/charms/istio_beacon_k8s/v0/service_mesh.py
dstathis: We unfortunately need to keep this as it is because of Juju limitations.

lib/charms/istio_beacon_k8s/v0/service_mesh.py
dstathis: This has been resolved by always keeping all keys in the ConfigMap so there is no need for the two commands to be one transaction.

PR #25, opened 2024-10-22
feat: add 'hardened' config value

ca-scribner:
This adds a config value 'hardened' which, when enabled, means the charm will create AuthorizationPolicies for any related application that requests them via the service-mesh relation


ca-scribner:
fwiw, I don't think we should land this.  It doesn't feel very 'hardened' to me - at best it is incomplete.

ca-scribner:
Closed in favor of #28 and some future efforts

PR #28, opened 2024-11-13
feat: name `AuthorizationPolicies` deterministically

### LINKED ISSUE #22: improve naming of `AuthorizationPolicies`
### Enhancement Proposal

Each related app can define multiple policies that apply to the same source/target pair.  At present, we use a naive naming convention that suffixes this relation's databag index to the end of the name, but this is not great because it will result in unnecessary churn of policies (ex: if a policy is added to the start of the databag, all other policies must be deleted and recreated).  

Naming these with just source_app_name-target_app_name isn't enough either, is it?  I think we need to support multiple policies per source/target pair?

Maybe we can use a hash as a suffix, as suggested [here](https://github.com/canonical/istio-beacon-k8s-operator/pull/12#discussion_r1796177783)?

ca-scribner:
## Issue
See #22

## Solution
This commit implements a less naive naming convention which:
* replaces the suffix with a hash value derived from the content of the policy
* adds checks to avoid creating a name that is longer than the Kubernetes max resource name length, truncating the name if required

## Context
#22

## Testing Instructions
Everything is covered by tests.

An optional manual test could be executing the test suite with `--keep-models` and inspecting the resulting Authorization Policies.

## Upgrade Notes
-

Closes #22

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -414,7 +414,7 @@ class IstioBeaconCharm(ops.CharmBase):
         """Generate a unique name for an AuthorizationPolicy, suffixing a hash of the MeshPolicy to avoid collisions.
 
         The name has the following general format:
-            {app_name}-{model_name}-policy-{source_app_name}-{target_app_name}-{hash}
+            {app_name}-{model_name}-policy-{source_app_name}-{source_namespace}-{target_app_name}-{hash}
         but source_app_name and target_app_name will be truncated if the total name exceeds Kubernetes's limit of 253
         characters.
         """
--- a/src/charm.py
+++ b/src/charm.py
@@ -5,12 +5,14 @@
 
 """Istio Beacon Charm."""
 
+import hashlib
 import logging
 import time
-from typing import Dict
+from typing import Dict, List
 
 import ops
-from charms.istio_beacon_k8s.v0.service_mesh import ServiceMeshProvider
+import pydantic
+from charms.istio_beacon_k8s.v0.service_mesh import MeshPolicy, ServiceMeshProvider
 from charms.prometheus_k8s.v0.prometheus_scrape import MetricsEndpointProvider
 from lightkube.core.client import Client
 from lightkube.core.exceptions import ApiError
@@ -65,7 +67,7 @@ class IstioBeaconCharm(ops.CharmBase):
 
         self._lightkube_field_manager: str = self.app.name
         self._lightkube_client = None
-        self._managed_labels = f"{self.app.name}-{self.model.name}"
+        self._app_identity = f"{self.app.name}-{self.model.name}"
 
         self._telemetry_labels = {
             f"charms.canonical.com/{self.model.name}.{self.app.name}.telemetry": "aggregated"
@@ -217,7 +219,7 @@ class IstioBeaconCharm(ops.CharmBase):
 
         self.unit.status = ActiveStatus()
 
-    def _build_authorization_policies(self, mesh_info):
+    def _build_authorization_policies(self, mesh_info: List[MeshPolicy]):
         """Build authorization policies for all related applications."""
         authorization_policies = [None] * len(mesh_info)
         for i, policy in enumerate(mesh_info):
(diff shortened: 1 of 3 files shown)


src/charm.py
```diff
@@ -407,6 +410,45 @@ def mesh_labels(self):
             "istio.io/use-waypoint-namespace": self.model.name,
         }
 
+    def _generate_authorization_policy_name(self, mesh_policy: MeshPolicy) -> str:
+        """Generate a unique name for an AuthorizationPolicy, suffixing a hash of the MeshPolicy to avoid collisions.
+
+        The name has the following general format:
+            {app_name}-{model_name}-policy-{source_app_name}-{target_app_name}-{hash}
```
IbraAoad, line 417:
```suggestion
            {app_name}-{model_name}-policy-{source_app_name}-{source_namespace}-{target_app_name}-{hash}
```

tests/unit/test_charm.py
```diff
@@ -229,3 +230,78 @@ def test_sync_waypoint_resources_remove_labels(harness: Harness[IstioBeaconCharm
...
+)
+def test_generate_authorization_policy_name(
+    beacon_name, beacon_namespace, mesh_policy, expected_name, harness: Harness[IstioBeaconCharm]
+):
+    """Test the _generate_authorization_policy_name() method."""
+    # TODO: how do we set app name in harness?  For now, just use the default
+    harness.set_model_name(beacon_namespace)
+    harness.begin()
+    charm = harness.charm
+
```
IbraAoad, line 304:
```suggestion
charm.model.app.name = beacon_name
```

tests/unit/test_charm.py
```diff
@@ -229,3 +230,78 @@ def test_sync_waypoint_resources_remove_labels(harness: Harness[IstioBeaconCharm
...
+                target_app_name="targetApp",
+                target_namespace="targetNamespace",
+                target_service=None,
+                endpoints=[
+                    Endpoint(hosts=["host1"], ports=[80], methods=["GET"], paths=["/path1"])
+                ],
+            ),
+            # Note: if this test fails because the hash has changed, that means upgrading from a previous version to
+            # this one will result in a delete/recreate of all policies.  Decide if that is acceptable.
+            "istio-beacon-k8s-beaconNamespace-policy-senderApp-senderNamespace-targetApp-65994572",
```
IbraAoad, line 254: Can we change this to be `beaconApp` to match the name or remove it all together

tests/unit/test_charm.py
```diff
@@ -229,3 +230,78 @@ def test_sync_waypoint_resources_remove_labels(harness: Harness[IstioBeaconCharm
...
+                target_namespace="targetNamespace",
+                target_service="my-service",
+                endpoints=[
+                    Endpoint(hosts=["host1"], ports=[80], methods=["GET"], paths=["/path1"]),
+                    Endpoint(hosts=["host2"], ports=[80], methods=["GET"], paths=["/path1"]),
+                ],
+            ),
+            # Note: if this test fails because the hash has changed, that means upgrading from a previous version to
+            # this one will result in a delete/recreate of all policies.  Decide if that is acceptable.
+            "istio-beacon-k8s-beaconNamespace-policy-senderApp-senderNamespace-targetApp-434ce4ee",
```
IbraAoad, line 273: Same here

tests/unit/test_charm.py
```diff
@@ -229,3 +230,78 @@ def test_sync_waypoint_resources_remove_labels(harness: Harness[IstioBeaconCharm
...
+                target_namespace="targetNamespace678901234567890123456789012345678901234567890123",
+                target_service="my-service",
+                endpoints=[
+                    Endpoint(hosts=["host1"], ports=[80], methods=["GET"], paths=["/path1"]),
+                    Endpoint(hosts=["host2"], ports=[80], methods=["GET"], paths=["/path1"]),
+                ],
+            ),
+            # Note: if this test fails because the hash has changed, that means upgrading from a previous version to
+            # this one will result in a delete/recreate of all policies.  Decide if that is acceptable.
+            "istio-beacon-k8s-beaconNamespace678901234567890123456789012345678901234567890123-policy-senderApp012345678901234567890-senderNamespace678901234567890-targetApp012345678901234567890-506b5a7c",
```
IbraAoad, line 292: here as well

IbraAoad (approved):

tests/unit/test_charm.py
ca-scribner: agreed.  added as part of another commit

PR #29, opened 2024-11-14
feat: add config to control if Authorization Policies are generated

ca-scribner:
## Issue

Previously, all applications requesting Authorization Policies via the service-mesh relation would receive those policies.  Because the presence of any Authorization Policy for a target results in all traffic to that target requiring an Authorization Policy, this could have a detrimental effect on an environment where not all traffic is being described by Authorization Policies.

## Solution
To mitigate this, this commit adds a toggle for Authorization Policy generation, leaving that feature on by default.  If disabled, Authorization Policies will not be created for any related application, even if they define Policies in their service-mesh relation data.

## Context
-

## Testing Instructions
Unit tests cover this change.

To manually test, do:
```
tox -e integration -- --keep-models

# confirm authorization policy exists
kubectl get authorizationpolicies.security.istio.io -A
# should have 1 AP

# update config to disable AP generation
juju config istio-beacon-k8s create-authorization-policies=false

kubectl get authorizationpolicies.security.istio.io -A
# should have 0 APs
```

## Upgrade Notes
-

### DIFF
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -56,6 +56,19 @@ parts:
 
 config:
   options:
+    manage-authorization-policies:
+      type: boolean
+      default: true
+      description: >
+        Automatically create Istio authorization policies for any charm requiresting them over the service mesh 
+        relation.  If set to false, the charm will not create any authorization policies, but will still do other 
+        functions like provide a waypoint.
+    model-on-mesh:
+      type: boolean
+      default: false
+      description: >
+        Add this charm's model to the service mesh. 
+        All charms in this model will automatically be added to the mesh.
     ready-timeout:
       type: int
       default: 100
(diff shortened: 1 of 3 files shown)


charmcraft.yaml
```diff
@@ -56,6 +56,19 @@ parts:
 
 config:
   options:
+    create-authorization-policies:
```
IbraAoad, line 59: tbh I like this explicit naming of the config better than `managed-mode`, tagging the team just to make sure we're all aligned @dstathis @simskij

charmcraft.yaml
```diff
@@ -56,6 +56,19 @@ parts:
...
+      type: boolean
+      default: true
+      description: >
+        Automatically create Istio authorization policies for any charm requiresting them over the service mesh
```
IbraAoad, line 63:
```suggestion
        Automatically create Istio authorization policies for any charm requesting them over the service mesh 
```
Feel free to ignore the suggestion if that was actually a word and my English failed me :D

IbraAoad (approved):

charmcraft.yaml
ca-scribner: ty for flagging this, I meant to do that myself and forgot.  this name ended up really verbose and literal, but clearer imo.  I'm flexible on it though if others disagree

charmcraft.yaml
ca-scribner: Your English served you better than mine here :)

charmcraft.yaml
ca-scribner: Renamed to `manage-authorization-policies` based on an offline discussion

PR #32, opened 2024-11-20
Charm tracing

PietroPasotti:
this PR adds charm-tracing to the istio-beacon charm

tested:

- [x] on a demo deployment

image

Known limitations: 
it will break if tempo is related with a tls provider.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -64,6 +64,7 @@ WAYPOINT_RESOURCE_TYPES = {RESOURCE_TYPES["Gateway"]}
 @trace_charm(
     tracing_endpoint="_charm_tracing_endpoint",
     # we don't add a cert because istio does TLS his way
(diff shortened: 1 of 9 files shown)


src/charm.py
```diff
@@ -59,6 +61,11 @@
 WAYPOINT_RESOURCE_TYPES = {RESOURCE_TYPES["Gateway"]}
 
 
+@trace_charm(
+    tracing_endpoint="_charm_tracing_endpoint",
+    # we don't add a cert because istio does TLS his way
```
IbraAoad, line 66:
```suggestion
    # we don't add a cert because istio does TLS his way
    # TODO: fix this when https://github.com/canonical/istio-beacon-k8s-operator/issues/33 is closed
```

IbraAoad (approved):

PR #46, opened 2024-12-20
Modeloperator traffic

### LINKED ISSUE #39: Enabling `model-on-mesh` blocks traffic from `kubernetes` to the model's operator
### Bug Description

@wood-push-melon reported that, after turning on the `model-on-mesh` config, he saw the following ztunnel logs:

```
2024-11-28T02:45:11.030584Z    error    access    connection complete    src.addr=172.31.20.48:3348 src.workload="kubernetes" src.namespace="default" dst.addr=10.1.48.89:17071 dst.service="modeloperator.testing.svc.cluster.local" dst.workload="modeloperator-797f7c5889-9nn7g" dst.namespace="testing" direction="inbound" bytes_sent=0 bytes_recv=0 duration="0ms" error="connection closed due to policy rejection"
```

Specifically, we see "connection closed due to policy rejection" so I think that means reproduction requires us to have a policy bound at ztunnel to the model operator (maybe there was a global allow-nothing policy in place in @wood-push-melon's env?)

Not sure what communication that would be or whether it is important.  Could it be a health check?  We'd need to investigate further to triage. 

### To Reproduce

* deploy beacon to a model
* enable model-on-mesh
* add an authorization policy
* watch the logs in the ztunnel pod to see if anything gets blocked

### Environment

-

### Relevant log output

```shell
-
```

### Additional context

_No response_

dstathis:
## Issue
Fixes #39

## Solution
When `model-on-mesh` is enabled, create an `AuthorizationPolicy` to allow traffic to the modeloperator.

## Testing Instructions
* Deploy istio-k8s and istio-beacon-k8s
* set model-on-mesh=true
* check authorization policies
* observe ztunnel logs have no errors

## Upgrade Notes
Fixed an issue where the juju controller could not reach the modeloperator when model-on-mesh was enabled.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -298,7 +298,7 @@ class IstioBeaconCharm(ops.CharmBase):
         # We need to allow the juju controller to be able to talk to the model operator
         if self.config["model-on-mesh"]:
             authorization_policies.append(
-                    RESOURCE_TYPES["AuthorizationPolicy"](  # type: ignore
+                RESOURCE_TYPES["AuthorizationPolicy"](  # type: ignore
                     metadata=ObjectMeta(
                         name=f"allow-traffic-to-{self.model.name}-modeloperator",
                         namespace=self.model.name,
--- a/src/charm.py
+++ b/src/charm.py
@@ -236,7 +236,7 @@ class IstioBeaconCharm(ops.CharmBase):
         self.unit.status = ActiveStatus()
 
     def _build_authorization_policies(self, mesh_info: List[MeshPolicy]):
-        """Build authorization policies for all related applications."""
+        """Build all managed authorization policies."""
         authorization_policies = [None] * len(mesh_info)
         for i, policy in enumerate(mesh_info):
             target_service = policy.target_service or policy.target_app_name
--- a/src/charm.py
+++ b/src/charm.py
@@ -300,7 +300,7 @@ class IstioBeaconCharm(ops.CharmBase):
             authorization_policies.append(
                 RESOURCE_TYPES["AuthorizationPolicy"](  # type: ignore
                     metadata=ObjectMeta(
-                        name=f"allow-traffic-to-{self.model.name}-modeloperator",
+                        name=f"{self.app.name}-{self.model.name}-policy-all-sources-modeloperator",
                         namespace=self.model.name,
                     ),
                     spec=AuthorizationPolicySpec(
--- a/src/charm.py
+++ b/src/charm.py
@@ -39,6 +39,7 @@ from models import (
     Rule,
     Source,
     To,
+    WorkloadSelector,
 )
 
 logger = logging.getLogger(__name__)
@@ -293,6 +294,24 @@ class IstioBeaconCharm(ops.CharmBase):
                     # exclude_none=True because null values in this data always mean the Kubernetes default
                 ).model_dump(by_alias=True, exclude_unset=True, exclude_none=True),
             )
+
+        # We need to allow the juju controller to be able to talk to the model operator
+        if self.config["model-on-mesh"]:
+            authorization_policies.append(
+                    RESOURCE_TYPES["AuthorizationPolicy"](  # type: ignore
+                    metadata=ObjectMeta(
+                        name=f"allow-traffic-to-{self.model.name}-modeloperator",
+                        namespace=self.model.name,
+                    ),
+                    spec=AuthorizationPolicySpec(
+                        selector=WorkloadSelector(
+                            matchLabels={"operator.juju.is/name": "modeloperator"}
+                        ),
+                        rules=[Rule()],
+                    ).model_dump(by_alias=True, exclude_unset=True, exclude_none=True),
+                )
+            )
+
         return authorization_policies
 
     def _construct_waypoint(self):
--- a/src/models.py
+++ b/src/models.py
@@ -70,7 +70,7 @@ class PolicyTargetReference(BaseModel):
 
 
 class WorkloadSelector(BaseModel):
-    """WorkloadSelector defines the selector for the policy."""
+    """WorkloadSelector defines the target of the policy for ztunnel bound policies."""
 
     matchLabels: Dict[str, str]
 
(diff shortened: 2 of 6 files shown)


tests/integration/test_charm.py
ca-scribner, line 1: I had trouble actually proving the traffic is being allowed to reach modeloperator.  If I do `tox -e integration -- --keep-models`, then deploy my own pod with curl and try to curl the model operator in beacon's namespace, I see no policy rejections but I do see timeouts after 10s.  This is different from when I curl a model operator in a different namespace, which will try to do a tls handshake right away.  I'm not sure what is going wrong here

ca-scribner (commented):
the direction on this lgtm.  I had a few minor suggestions, plus had trouble proving traffic was allowed.

tests/integration/test_charm.py
```diff
@@ -70,12 +71,15 @@ async def test_mesh_config(ops_test: OpsTest):
         [APP_NAME], status="active", timeout=1000, raise_on_error=False
     )
     await validate_labels(ops_test, APP_NAME, should_be_present=True)
+    validate_policy_exists(ops_test, f"allow-traffic-to-{ops_test.model.name}-modeloperator")
```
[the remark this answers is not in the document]
ca-scribner: Thought about this a bit more today after we talked, and I still think the right integration test here is to assert that traffic from off mesh can reach the model operator even if the model is put on the mesh.  

If we decide to just validate that a policy exists, we should move this test to the unit tests.  We can confirm we attempted to create the policy in a unit test and be just as certain of the validity.  The only things the unit test would miss would be if we modeled the object's schema incorrectly, and that will still raise integration test errors since the kubernetes API will reject us when we try to create the objects

src/charm.py
```diff
@@ -293,6 +294,24 @@ def _build_authorization_policies(self, mesh_info: List[MeshPolicy]):
                     # exclude_none=True because null values in this data always mean the Kubernetes default
                 ).model_dump(by_alias=True, exclude_unset=True, exclude_none=True),
             )
+
+        # We need to allow the juju controller to be able to talk to the model operator
+        if self.config["model-on-mesh"]:
+            authorization_policies.append(
+                RESOURCE_TYPES["AuthorizationPolicy"](  # type: ignore
+                    metadata=ObjectMeta(
+                        name=f"allow-traffic-to-{self.model.name}-modeloperator",
```
ca-scribner, line 303: Can we name this similar to the other [authorization policies this app creates](https://github.com/canonical/istio-beacon-k8s-operator/blob/c96497ef0563b72c4d87207b0d0ffbebd67d723f/src/charm.py#L459C13-L459C105)?  They use:
> {app_name}-{model_name}-policy-{source_app_name}-{source_namespace}-{target_app_name}-{hash}

some of that doesn't apply, but maybe we do
```
{app_name}-{model_name}-policy-all-sources-modeloperator-{hash}
```
?

I'm not particular, just something that feels like they came from the same application

src/charm.py
```diff
@@ -293,6 +294,24 @@ def _build_authorization_policies(self, mesh_info: List[MeshPolicy]):
                     # exclude_none=True because null values in this data always mean the Kubernetes default
                 ).model_dump(by_alias=True, exclude_unset=True, exclude_none=True),
             )
+
+        # We need to allow the juju controller to be able to talk to the model operator
```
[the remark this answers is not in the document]
dstathis: Done.

src/models.py
```diff
@@ -61,7 +61,7 @@ class Action(str, Enum):
 
 
 class PolicyTargetReference(BaseModel):
-    """PolicyTargetReference defines the target of the policy."""
+    """PolicyTargetReference defines the target of the policy for waypoint bound policies."""
```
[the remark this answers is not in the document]
dstathis: Done.

src/models.py
```diff
@@ -130,8 +130,15 @@ class AuthorizationPolicySpec(BaseModel):
...
-    # selector: WorkloadSelector
-    # provider
+
+    @model_validator(mode="after")
+    def validate_target(self):
+        """Validate that exactly one of targetRefs and selector is defined."""
+        if (self.targetRefs is None and self.selector is None) or (
+            self.targetRefs is not None and self.selector is not None
+        ):
+            raise ValueError("Only one of targetRefs and selector can be set")
```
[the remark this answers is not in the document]
dstathis: Done.

tests/integration/test_charm.py
dstathis: If you reproduce the original issue, you can then refresh to this branch.

ca-scribner: afaict this was due to [this bug](https://github.com/istio/istio/issues/54630) along with the specific procedure I used to test.  It generally wont be an issue, and has been fixed in full for later istio versions.

tests/integration/test_charm.py
```diff
@@ -130,6 +138,24 @@ async def test_service_mesh_relation(ops_test: OpsTest, service_mesh_tester):
     )
 
 
+@pytest.mark.abort_on_fail
+async def test_modeloperator_rule(ops_test: OpsTest, service_mesh_tester):
+    # Ensure model is on mesh
+    await ops_test.model.applications[APP_NAME].set_config({"model-on-mesh": "true"})
+    await ops_test.track_model("off-mesh-model")
```
ca-scribner, line 145: Not required, but a suggestion to make `juju models` more readable
```suggestion
    await ops_test.track_model("off-mesh-model", model_name=f"{ops_test.model.name}-off-mesh-model")
```

If you do accept this, we should also change `ops_test.track_model("istio-system")` above to `    await ops_test.track_model("istio-system", model_name=f"{ops_test.model.name}-istio-system")`.  

Making these two changes imo makes the `juju models` return more readable when debugging:

```
juju models


Controller: local-microk8s

Model                           Cloud/Region              Type        Status     Units  Access  Last connection
controller                      local-microk8s/localhost  kubernetes  available  1       admin  just now
test-charm-uu4k                 local-microk8s/localhost  kubernetes  available  4       admin  8 seconds ago
test-charm-uu4k-istio-system    local-microk8s/localhost  kubernetes  available  1       admin  6 minutes ago
test-charm-uu4k-off-mesh-model  local-microk8s/localhost  kubernetes  available  -       admin  7 seconds ago
```

dstathis: Yes this is awesome!

PR #79, opened 2025-06-09
fix: correctly scope AuthorizationPolicy for modeloperator

### LINKED ISSUE #78: istio-beacon's authorizationpolicy for the modeloperator does not work
### Bug Description

In #46 we added an `AuthorizationPolicy` that allowed all traffic to the modeloperator in beacon's model.  This looked roughly like:

```yaml
apiVersion: security.istio.io/v1
kind: AuthorizationPolicy
metadata:
  labels:
    app.juju.is/created-by: istio-beacon-k8s
    app.kubernetes.io/instance: istio-beacon-k8s-test-charm-bcdk
    kubernetes-resource-handler-scope: istio-authorization-policy
  name: istio-beacon-k8s-test-charm-bcdk-policy-all-sources-modeloperator
  namespace: test-charm-bcdk
spec:
  rules:
  - {}
  selector:
    matchLabels:
      operator.juju.is/name: modeloperator
```

This targets the modeloperator workload via a workload selector.  

Currently in Juju 3.6.6, our `test_modeloperator_rule` CI which asserts this is working is failing.  I see the modeloperator pod does not have the `{operator.juju.is/name: modeloperator}` label anymore, and instead has `{juju-modeloperator: modeloperator}`.  This means the AuthorizationPolicy does nothing, and beacon's policies block traffic to the modeloperator.  This is confirmed by looking at the ztunnel logs, which show communication being rejected by absence of an allow policy.

I'm not sure when this change in Juju occurred, but I'm guessing in a recent-ish Juju (3.6.x?) this label has been changed.  

We should do a few things to address this:
* fix this bug.  should we support the old labeling scheme, or just assume people have a new-enough Juju to have the new scheme?
* make an issue/proposal on what we should change to catch this proactively.  We shouldn't notice this by accident after Juju has released it.  Since we're relying on implementation details in juju like this label, we may need more proactive testing of pre-release candidates to catch any regressions early
  * ✅ https://github.com/canonical/istio-k8s-operator/issues/70

### To Reproduce

Run the CI in juju 3.6.6

### Environment

juju 3.6.6

### Relevant log output

```shell
-
```

### Additional context

_No response_

ca-scribner:
## Issue
Fixes #78

## Solution
Updates the label used to scope the AuthorizationPolicy for the modeloperator in beacon's namespace to match the current Juju naming convention.

An alternate approach here could have been to create an AP for both the old and new schema, or maybe detecting which is appropriate.  We can do that, but since we require Juju 3.6 before this goes stable and all the 3.6's will follow this schema, that feels unnecessary(?).

## Context
#78

## Testing Instructions
`test_modeloperator_rule` integration test covers this

## Upgrade Notes
None

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -304,7 +304,7 @@ class IstioBeaconCharm(ops.CharmBase):
                     ),
                     spec=AuthorizationPolicySpec(
                         selector=WorkloadSelector(
-                            matchLabels={"operator.juju.is/name": "modeloperator"}
+                            matchLabels={"juju-modeloperator": "modeloperator"}
                         ),
                         rules=[Rule()],
                     ).model_dump(by_alias=True, exclude_unset=True, exclude_none=True),


src/charm.py
```diff
@@ -304,7 +304,7 @@ def _build_authorization_policies(self, mesh_info: List[MeshPolicy]):
                     ),
                     spec=AuthorizationPolicySpec(
                         selector=WorkloadSelector(
-                            matchLabels={"operator.juju.is/name": "modeloperator"}
+                            matchLabels={"juju-modeloperator": "modeloperator"}
```
ca-scribner, line 307: Highlighting this so reviewers don't miss it:

The way this is done is just changing beacon to use the new schema.  That means that beacon works on new juju (juju >=3.6.something), but no longer works for older Jujus (maybe <3.6?  not sure).  Its the simplest solution, but also means we dont work on older Jujus.  That said, we also have [juju >= 3.6](https://github.com/canonical/istio-beacon-k8s-operator/blob/9ab26a497246bbe51b2d9c1efaa185ce4e24327c/charmcraft.yaml#L9) set in the assumes directive, so I argue that's ok.

...but if anyone else disagrees, we can do this more backward compatible.  If we choose to be more backward compatible though, I propose we merge this pr now (since this issue breaks the CI right now and blocks all other development in Beacon) and we can do a more nuanced fix in a follow-on PR.

Issue #34, opened 2024-11-25
AuthorizationPolicy does not work after adding a charm to the service mesh


wood-push-melon:
### Bug Description

After adding a requirer charm to the service mesh, the AuthorizationPolicy that comes with the requirer charm does not apply because:

- AuthorizationPolicy attaches to the requirer's k8s svc of the charm application
- Only the requirer charm's application workloads are labeled to use the Waypoint, not the requirer's k8s svc

Therefore the AuthorizationPolicy created through the `service-mesh` integration will not simply apply. Please check the following reproduce steps.

When adding a single requirer charm to the service mesh, the istio-beacon-k8s charm is supposed also to label the corresponding k8s svc.

### To Reproduce

Assume we've deployed the istio control plane charm `istio-k8s` in the `istio-system` namespace. 

```
# Add a juju model. The following steps are all operated in the `test` juju model
juju add-model test

# Deploy the istio-beacon charm
juju deploy istio-beacon-k8s --channel edge

# Deploy two application charms A & B. Charm A and B have a certain juju integration declared
juju deploy A
juju deploy B
juju integrate A B

# Add both application charms to the service mesh. This will create an AuthorizationPolicy targeted for application A's k8s svc. This policy will enforce certain L7 traffic initiated from B to A.
juju integrate A:service-mesh istio-beacon-k8s
juju integrate B:service-mesh istio-beacon-k8s

# Get into application B to test if AuthorizationPolicy works.
k -n test exec -it B -- /bin/bash
curl -v http://A/{the enforced endpoint}
```

At this point, the workloads (pods) of A and B are properly labeled as following (an illustrative example):

```
istio.io/dataplane-mode: ambient
istio.io/use-waypoint: istio-beacon-k8s-test-waypoint
istio.io/use-waypoint-namespace: test
```

But the A's k8s svc is not labeled to use the Waypoint. This will cause the AuthorizationPolicy not to apply because the created AuthrozationPolicy is targeted to the A's k8s svc.

We can also check if the k8s svc is using the Waypoint properly using the following command:

```
istioctl ztunnel-config service
```

### Environment

- microk8s: v1.27.16
- istio-beacon-k8s: latest/edge

### Relevant log output

```shell
N/A
```

ca-scribner:
Ah I see what you're saying.  I think this is another one of the "Authorization Policies in ambient work differently than in sidecar" mistakes that we've been making.  

[...]

PR #86, opened 2025-06-16
fix: add istio mesh labels to service

adhityaravi:
WIP

### DIFF
--- a/lib/charms/istio_beacon_k8s/v0/service_mesh.py
+++ b/lib/charms/istio_beacon_k8s/v0/service_mesh.py
@@ -154,12 +154,12 @@ import pydantic
 from lightkube.core.client import Client
 from lightkube.models.meta_v1 import ObjectMeta
 from lightkube.resources.apps_v1 import StatefulSet
-from lightkube.resources.core_v1 import ConfigMap
+from lightkube.resources.core_v1 import ConfigMap, Service
 from ops import CharmBase, Object, Relation
 
 LIBID = "3f40cb7e3569454a92ac2541c5ca0a0c"  # Never change this
 LIBAPI = 0
-LIBPATCH = 2
+LIBPATCH = 3
 
 PYDEPS = ["lightkube", "pydantic"]
 
@@ -364,6 +364,8 @@ class ServiceMeshConsumer(Object):
     def _set_labels(self, labels: dict) -> None:
         client = Client(namespace=self._charm.model.name, field_manager=self._charm.app.name)
         stateful_set = client.get(res=StatefulSet, name=self._charm.app.name)
+        service = client.get(res=Service, name=self._charm.app.name)
+
         try:
             config_map = client.get(ConfigMap, self._label_configmap_name)
         except httpx.HTTPStatusError as e:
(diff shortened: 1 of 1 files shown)


ca-scribner:
I think this work closes #34, am I right?  I've added a note to that issue about follow-on work, but if I've got the wrong issue please help me out and put that note in the right place :)

lib/charms/istio_beacon_k8s/v0/service_mesh.py
```diff
@@ -344,11 +346,17 @@ def _set_labels(self, labels: dict) -> None:
...
         if stateful_set.spec:
             stateful_set.spec.template.metadata.labels.update(labels)  # type: ignore
+        if service.metadata:
+            service.metadata.labels = service.metadata.labels or {}
+            service.metadata.labels.update(labels)
         config_map.data = {"labels": json.dumps(labels)}
+
         client.patch(res=ConfigMap, name=self._label_configmap_name, obj=config_map)
         client.patch(res=StatefulSet, name=self._charm.app.name, obj=stateful_set)
+        client.patch(res=Service, name=self._charm.app.name, obj=service)
```
ca-scribner, line 359: The PR description mentions "also some additional labels for ease of solution" which I think is aluding to the service technically just needs a subset of the labels here, but we add them all because its simpler.  Can you update the PR description to just be more explicit about that?  I think if we read back and see that PR desc, we're going to say "oh yeah there was something funny we did here... what was that?"

I'm thinking something like:

> While subscribing a Pod or Namespace to the service mesh requires the labels `{...}`, strictly speaking we only need to label a Service with `{...}` to have its traffic captured by a waypoint.  But since the Service labels are a subset of the Pod/Namespace labels and there's no cost to adding extra labels, we just apply the same label set to all objects.

I'd also love to see a brief comment right here saying something like "# We apply all labels to the Service even though not all are necessary.  See https://github.com/canonical/istio-beacon-k8s-operator/pull/86/ for more explanation", but if you or others disagree then we can omit the comment.

tests/integration/testers/service-mesh-tester/charmcraft.yaml
```diff
@@ -13,6 +13,14 @@ bases:
         channel: "20.04"
 
 config:
+  options:
+    auto-join-mesh:
+      type: boolean
+      default: true
+      description: |
+        Automatically join the mesh by applying relevant labels to the charm's K8s resources on 
+        relating via the service_mesh interface
```
ca-scribner, line 22: Since this is just a tester for our own use, we can be more direct about what this is for in this charm.  What about:

```suggestion
      description: |
        Used for `ServiceMeshConsumer`'s `auto_join` argument, which controls whether this charm will add itself to the mesh if there is a service_mesh relation established to a beacon.
```

Feel free to use your own words though :)

tests/integration/test_charm.py
```diff
@@ -77,29 +77,66 @@ async def test_mesh_config(ops_test: OpsTest):
...
+            "sender2",
+        ],
+        raise_on_error=False,
+    )
 
-    await ops_test.model.wait_for_idle([APP_NAME, "receiver1", "sender1", "sender2"])
+
+@pytest.mark.abort_on_fail
+@pytest.mark.parametrize("model_on_mesh", [True, False])
+async def test_service_mesh_relation(ops_test: OpsTest, model_on_mesh):
```
ca-scribner, line 121: I realize these didn't have docstrings before you were here, but lets start adding them so its clear for the next dev

Something like this?  Feel free to change

```suggestion
async def test_service_mesh_relation(ops_test: OpsTest, model_on_mesh):
    """Test that the service mesh relation correctly puts applications on the mesh and opens traffic as expected.
    
    The test sets `auto-join-mesh` for the tester charm based on the `model_on_mesh` parameter.  So:
    * when `model_on_mesh=True` we set `auto-join-mesh=False` to test that the model has subscribed the apps
    * when `model_on_mesh=False` we set `auto-join-mesh=True` to test that the apps have subscribed themselves
    """
```

ca-scribner (approved):
lgtm, with some suggestions

lib/charms/istio_beacon_k8s/v0/service_mesh.py
adhityaravi: Yes, totally agree about the description. I have added it. I am not sure about the comment, I think we can use git blame for that. and find the PR desc.

tests/integration/test_charm.py
adhityaravi: done! added docstrings to the test

PR #88, opened 2025-06-19
feat: add istio-beacon charm to the service mesh

ca-scribner:
## Issue
Previously, istio-beacon had no direct way of adding itself to the service mesh.  This meant that if model-on-mesh=False, the istio-beacon charm would not be on the mesh.  This was raised in #41.

## Solution

This commit fixes this by adding logic to always label istio-beacon's resources (Pods, Service) to put it on the service mesh and be captured by the waypoint it deploys.

## Context
#41

## Testing Instructions

Istio adds the annotation `ambient.istio.io/redirection=enabled` to any Pod on the mesh.  An additional integration test has been added here to take advantage of that and confirm beacon's pod is on the mesh.

To test manually, you can also inspect the labels on the beacon Service and Pod - there should be istio enrollment labels present.

## Upgrade Notes
None

# Other notes
* Merge after #87


adhityaravi (approved):
LGTM!

src/charm.py
```diff
@@ -466,10 +483,14 @@ def _remove_labels(self):
...
-    def mesh_labels(self):
-        """Labels required for a workload to join the mesh."""
+    def mesh_labels_for_service_mesh_relation(self):
+        """Labels required for a workload to join the mesh, guarded by whether model-on-mesh=True."""
         if self.config["model-on-mesh"]:
             return {}
+        return self.mesh_labels()
+
+    def mesh_labels(self):
+        """Labels required for a workload to join the mesh."""
```
adhityaravi, line 493:
```suggestion
        """Labels required for a Charm's Kubernetes objects to join the mesh."""
```

Issue #15, opened 2024-10-03
Beacon does not behave properly when scaled.


dstathis:
### Bug Description

When Beacon is scaled beyond 1, It goes in to blocked state. It should instead continue to work and it should scale custom deployments accordingly.

### To Reproduce

`juju scale-application beacon 3`

### Environment

-

### Relevant log output

```shell
Unit                  Workload  Agent  Address     Ports  Message
istio-beacon-k8s/0*   active    idle   10.1.20.46         
istio-beacon-k8s/1    blocked   idle   10.1.20.3          Waypoint can only be provided on the leader unit.
```

### Note

- [ ] Make sure #21 is also fixed with this.

IbraAoad:
> Decisions from backlog refinement: 
> * In the near-term, we will keep creating a Gateway object and letting Istio "implement" the gateway from that object.  To control scaling, we will use a HorizontalPodAutoscaler as suggested [in the istio docs](https://istio.io/latest/docs/tasks/traffic-management/ingress/gateway-api/#automated-deployment)
> * For the longer term, we will do a spike to look at whether we should manually deploy and scale these, ideally using a sidecar for every gateway instance.   #64 

 _Originally posted by @ca-scribner in [#16](https://github.com/canonical/istio-ingress-k8s-operator/issues/16#issuecomment-2821603010)_

On top of the above the implementation should follow a global leader guard on all charm operations similar to what's implemented in https://github.com/canonical/istio-ingress-k8s-operator/pull/61

dstathis:
Fixed by #95.

PR #95, opened 2025-06-27
fix: scale beacon without going into Blocked state

adhityaravi:
## Issue
Beacon's replicas goes into a blocked state when the charm is scaled to more than 1 as described in issue #15. 

## Solution
Following the [solution](https://github.com/canonical/istio-ingress-k8s-operator/pull/61) for the `istio-ingress-k8s` charm: 

- Move the non-leader charm units into Active status with a message saying they are waiting to be a leader
- Add a [HorizontalPodAutoscaler](https://istio.io/latest/docs/tasks/traffic-management/ingress/gateway-api/#automated-deployment) to scale the waypoint pods automatically along with the charms
- Put all the charm operations behind a leader guard
- Add scenario and integration tests

## Context
The removal of the application resources specifically is not put behind a leader guard following this [conversation](https://github.com/canonical/istio-ingress-k8s-operator/pull/61#discussion_r2056539971)

## Testing Instructions
```bash
juju add-model istio-system
juju deploy istio-k8s --trust --channel=2/edge

# clone the source branch
charmcraft pack
juju deploy ./istio-beacon-k8s_ubuntu@22.04-amd64.charm --trust --resource metrics-proxy-image=docker.io/ibraaoad/metrics-proxy:v0.1.0
juju scale-application istio-beacon-k8s 3
juju status --watch 2s

juju scale-application istio-beacon-k8s 1
```

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -25,11 +25,16 @@ from charms.tempo_coordinator_k8s.v0.tracing import TracingEndpointRequirer
 from lightkube import Client
 from lightkube.core.exceptions import ApiError
 from lightkube.generic_resource import create_namespaced_resource
+from lightkube.models.autoscaling_v2 import (
+    CrossVersionObjectReference,
+    HorizontalPodAutoscalerSpec,
+)
 from lightkube.models.meta_v1 import ObjectMeta
 from lightkube.resources.apps_v1 import Deployment
+from lightkube.resources.autoscaling_v2 import HorizontalPodAutoscaler
 from lightkube.resources.core_v1 import Namespace
 from lightkube_extensions.batch import KubernetesResourceManager, create_charm_default_labels
-from ops.model import ActiveStatus, BlockedStatus, MaintenanceStatus
+from ops.model import ActiveStatus, MaintenanceStatus
 from ops.pebble import ChangeError, Layer
 
 from models import (
@@ -65,7 +70,12 @@ RESOURCE_TYPES = {
 AUTHORIZATION_POLICY_LABEL = "istio-authorization-policy"
 AUTHORIZATION_POLICY_RESOURCE_TYPES = {RESOURCE_TYPES["AuthorizationPolicy"]}
 WAYPOINT_LABEL = "istio-waypoint"
-WAYPOINT_RESOURCE_TYPES = {RESOURCE_TYPES["Gateway"]}
+WAYPOINT_RESOURCE_TYPES = {
+    RESOURCE_TYPES["Gateway"],
+    HorizontalPodAutoscaler,
+}
+
+PEERS_RELATION = "peers"
 
 
 @trace_charm(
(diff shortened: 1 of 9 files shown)


src/charm.py
```diff
@@ -243,8 +270,15 @@ def _is_waypoint_ready(self) -> bool:
         return True
 
     def _sync_all_resources(self):
+        """Reconcile all resouces including gateway, horizontal pod scaler, and authorization policies.
+
+        * Add this beacon charm to the mesh by applying the required labels on the application resources
+        * Reconicle HPA and gateway resources to align replicas with unit count and ensure the readiness of waypoint deployment
+        * Set up the proxy service
+        * Build and reconcile the authorization policies
+        """
```
ca-scribner, line 279:
```suggestion
        """Reconcile all resources including gateway, horizontal pod autoscaler, and authorization policies.

This:
        * Adds this beacon charm to the mesh by applying the required labels on the application resources
        * Reconicles HPA and gateway resources to set the gateway replica count to be the same as the unit scale
        * Sets up the metrics proxy service
        * Builds and reconciles the authorization policies
        """
```

src/charm.py
```diff
@@ -377,6 +411,28 @@ def _construct_waypoint(self):
             spec=gateway.spec.model_dump(),
         )
 
+    def _construct_hpa(self, unit_count: int) -> HorizontalPodAutoscaler:
+        """Constructs a HorizontalPodAutoscaler resource targeting the waypoint Deployment.
+
+        This HPA is used to scale the waypoint workload automatically when the charm scales.
+        The scaling is achieved by setting the min and max replica settings in the HPA to be the same as the unit count. It is important to note that the HPA must target the waypoint's Deployment and not Gateway.
```
ca-scribner, line 418: something isn't right with the linting on this repo, but it should have complained about line width here...

```suggestion
        The scaling is achieved by setting the min and max replica settings in the HPA to be the same as the unit count.
        It is important to note that the HPA must target the waypoint's Deployment and not Gateway.
```

src/charm.py
```diff
@@ -377,6 +411,28 @@ def _construct_waypoint(self):
...
+        """
+        return HorizontalPodAutoscaler(
+            metadata=ObjectMeta(
+                name=self._waypoint_name,  # its a part of the resources related to waypoint, so use the waypoint name to id it
```
ca-scribner, line 422: and here

```suggestion
                # Identify it by the same name as the waypoint itself
                name=self._waypoint_name,
```

tests/integration/test_charm.py
```diff
@@ -22,12 +22,6 @@
 
 logger = logging.getLogger(__name__)
 
-METADATA = yaml.safe_load(Path("./charmcraft.yaml").read_text())
-APP_NAME = METADATA["name"]
-RESOURCES = {
-    "metrics-proxy-image": METADATA["resources"]["metrics-proxy-image"]["upstream-source"],
-}
-
```
ca-scribner, line 30: Fine for this PR, but in future split drivebys like these up into their own PR that is separate from the feature at hand.  No need to go crazy splitting it into a bunch of PRs (just one covering stuff like this and the additional docstrings you added is fine), but if I'm reviewing "scale beacon without going into blocked state" I hit stuff like this and wonder why it was joined to this PR

ca-scribner: Also, I get that probably you thought "I'm writing this test_charm_scaling.py and it could use these existing things".  Totally agree, but I still vote that refactor comes out since then its a really clear review of "I refactored the tests in a way that broke nothing"

tests/integration/test_charm_scaling.py
```diff
@@ -0,0 +1,129 @@
...
+        application_name="tester",
+        resources=resources,
+        trust=True,
+    )
+    await ops_test.model.wait_for_idle(
+        ["tester"],
+        status="active",
+        timeout=1000,
+    )
+    await ops_test.model.add_relation("tester:service-mesh", APP_NAME)
```
ca-scribner, line 74: Is this needed in the test?  I dont think its used

ca-scribner (approved):
A few minor nits or suggestions but otherwise lgtm!

Issue #76, opened 2025-06-02
Enable creating `AuthorizationPolicies` that bind to a workload not a Service (as in an L4 policy)


ca-scribner:
### Enhancement Proposal

# Issue

For every workload that istio-beacon's `Waypoint` serves, istio automatically creates a ztunnel-bound (L4) policy allowing `Waypoint`->`Workload`, which has the side effect of blocking any other traffic that goes directly to the Workloads (unless there are other L4 policies in place).  We then provide, through the `ServiceMeshConsumer` lib, a way for applications to request `AuthorizationPolicies` to allow traffic to a workload.  An example is:

```
self._mesh = ServiceMeshConsumer(
    self,
    policies=[
        Policy(
            relation="metrics",
#             service="some-service",  # Optional - defaults to the related app's default juju service
            endpoints=[
                Endpoint(
                    hosts=[self._my_host_name],
                    ports=[HTTP_LISTEN_PORT],
                    methods=["GET"],
                    paths=["/metrics"],
                ),
            ],
        ),
        Policy(
            relation="data",
            endpoint=[
                Endpoint(
                    hosts=[self._my_host_name],
                    ports[HTTP_LISTEN_PORT],
                    methods=["GET"]
                    paths=["/data"],
                ),
            ],
        ),
    ],
)
```

This application requests from beacon the generation of `AuthorizationPolicies` described by the above `Policy` objects.  These `AuthorizationPolicies` are **bound to a `Waypoint` and target a `Service`** (eg: they're "L7 policies" applied at the `Waypoint`).    As most networking in Kubernetes travels through a Service, this is usually the behaviour we want.

For prometheus(/grafana-agent/mimir), metrics collection is a pull action where prometheus scrapes the metrics endpoint **for each workload pod directly, bypassing the Service/Waypoint** (eg: prometheus does `curl POD_IP` rather than `curl SERVICE_URL`).  This is done because, if accessed through the service, Prometheus would get round-robined across the pods rather than hit each one deterministically.  In how we've currently modeled this, however, there's no way to enable this communication.  

# Solution

In general, to allow prometheus this direct-to-workload access we must create ztunnel-bound (L4) policies for this traffic.  This was confirmed by the upstream istio community [in this slack conversation](https://istio.slack.com/archives/C041EQL1XMY/p1748633158707859).  How we model/implement that in Juju needs to be figured out...

### Add a field to `Policy`

[...]

PR #103, opened 2025-08-05
Workload binding policies

adhityaravi:
# DRAFT: TBD

### DIFF
--- a/lib/charms/istio_beacon_k8s/v0/service_mesh.py
+++ b/lib/charms/istio_beacon_k8s/v0/service_mesh.py
@@ -147,7 +147,8 @@ for policy in self._mesh.mesh_info():
 import enum
 import json
 import logging
-from typing import Dict, List, Optional
+import warnings
+from typing import Dict, List, Literal, Optional, Union
 
 import httpx
 import pydantic
@@ -159,7 +160,7 @@ from ops import CharmBase, Object, RelationMapping
 
 LIBID = "3f40cb7e3569454a92ac2541c5ca0a0c"  # Never change this
 LIBAPI = 0
-LIBPATCH = 6
+LIBPATCH = 7
 
 PYDEPS = ["lightkube", "pydantic"]
 
@@ -193,12 +194,43 @@ class Endpoint(pydantic.BaseModel):
     paths: Optional[List[str]] = None
 
 
+class PolicyTargetType(str, enum.Enum):
+    """Target type for Policy classes."""
+
+    app = "app"
+    unit = "unit"
+
+
 class Policy(pydantic.BaseModel):
     """Data type for defining a policy for your charm."""
 
     relation: str
     endpoints: List[Endpoint]
     service: Optional[str] = None
+
+    def __init__(self, **data):
+        warnings.warn(
+            "Policy is deprecated. Use AppPolicy for fine-grained application-level policies "
+            "or UnitPolicy to allow access to charm units. For migration, Policy can be "
+            "directly replaced with AppPolicy.",
+            DeprecationWarning,
+            stacklevel=2
+        )
+        super().__init__(**data)
+
+
+class AppPolicy(pydantic.BaseModel):
+    """Data type for defining a policy for your charm application."""
+
+    relation: str
+    endpoints: List[Endpoint]
+    service: Optional[str] = None
+
+
+class UnitPolicy(pydantic.BaseModel):
+    """Data type for defining a policy for your charm unit."""
+
+    relation: str
 
 
 class MeshPolicy(pydantic.BaseModel):
@@ -209,6 +241,7 @@ class MeshPolicy(pydantic.BaseModel):
     target_app_name: str
     target_namespace: str
     target_service: Optional[str] = None
+    target_type: Literal[PolicyTargetType.app, PolicyTargetType.unit] = PolicyTargetType.app
     endpoints: List[Endpoint]
 
 
@@ -228,7 +261,7 @@ class ServiceMeshConsumer(Object):
         mesh_relation_name: str = "service-mesh",
         cross_model_mesh_requires_name: str = "require-cmr-mesh",
         cross_model_mesh_provides_name: str = "provide-cmr-mesh",
-        policies: Optional[List[Policy]] = None,
+        policies: Optional[List[Union[Policy, AppPolicy, UnitPolicy]]] = None,
         auto_join: bool = True,
     ):
         """Class used for joining a service mesh.
@@ -412,7 +445,7 @@ def build_mesh_policies(
         relation_mapping: RelationMapping,
         target_app_name: str,
         target_namespace: str,
-        policies: List[Policy],
+        policies: List[Union[Policy, AppPolicy, UnitPolicy]],
         cmr_application_data: Dict[str, CMRData]
 ) -> List[MeshPolicy]:
     """Generate MeshPolicy that implement the given policies for the currently related applications.
@@ -421,7 +454,7 @@ def build_mesh_policies(
         relation_mapping: Charm's RelatioMapping object, for example self.model.relations.
         target_app_name: The name of the target application, for example self.app.name.
         target_namespace: The namespace of the target application, for example self.model.name.
-        policies: List of Policy objects defining the access rules.
+        policies: List of Policy (deprecated), AppPolicy, or UnitPolicy objects defining the access rules.
         cmr_application_data: Data for cross-model relations, mapping app names to CMRData.
     """
     mesh_policies = []
@@ -436,16 +469,30 @@ def build_mesh_policies(
                 source_app_name = relation.app.name
                 source_namespace = target_namespace
 
-            mesh_policies.append(
-                MeshPolicy(
-                    source_app_name=source_app_name,
-                    source_namespace=source_namespace,
-                    target_app_name=target_app_name,
-                    target_namespace=target_namespace,
-                    target_service=policy.service,
-                    endpoints=policy.endpoints,
-                ).model_dump()
-            )
+            if isinstance(policy, UnitPolicy):
+                mesh_policies.append(
+                    MeshPolicy(
+                        source_app_name=source_app_name,
+                        source_namespace=source_namespace,
+                        target_app_name=target_app_name,
+                        target_namespace=target_namespace,
+                        target_service=None,
+                        target_type=PolicyTargetType.unit,
+                        endpoints=[],
+                    ).model_dump()
+                )
+            else:
+                mesh_policies.append(
+                    MeshPolicy(
+                        source_app_name=source_app_name,
+                        source_namespace=source_namespace,
+                        target_app_name=target_app_name,
+                        target_namespace=target_namespace,
+                        target_service=policy.service,
+                        target_type=PolicyTargetType.app,
+                        endpoints=policy.endpoints,
+                    ).model_dump()
+                )
     return mesh_policies
 
 
--- a/src/charm.py
+++ b/src/charm.py
@@ -15,6 +15,7 @@ import pydantic
 from charmed_service_mesh_helpers import charm_kubernetes_label
 from charms.istio_beacon_k8s.v0.service_mesh import (
     MeshPolicy,
+    PolicyTargetType,
     ServiceMeshProvider,
     label_configmap_name_template,
     reconcile_charm_labels,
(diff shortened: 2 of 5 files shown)


adhityaravi:
@ca-scribner @dstathis A couple of questions - 

- Currently i only **deprecated** the Policy class with a warning. Should this rather be completely removed? I didnt get what was decided and meant by deprecated in #76. But then I also saw this was supposed to be a breaking change and my current implementation does not make this a breaking change. 
- The UnitPolicy currently targets all the units of the juju application to which the UnitPolicy is added. This is correct, yes?
- Adding a UnitPolicy kind of gives the source app a free ticket to access all endpoints of the target workload. 
- Also, if the target app does not have any AppPolicy, UnitPolicy also lets the source app to access the target service without any restriction (as mentioned by @ca-scribner in #76 since we dont have that allow nothing policy for waypoints yet). Was that supposed to be part of this PR?

ca-scribner:
> @ca-scribner @dstathis A couple of questions and remarks -
> 
> * Currently i only **deprecated** the Policy class with a warning. Should this rather be completely removed? I didnt get what was decided and meant by deprecated in [Enable creating `AuthorizationPolicies` that bind to a workload not a Service (as in an L4 policy) #76](https://github.com/canonical/istio-beacon-k8s-operator/issues/76).
* we have the warning for now, but after we've landed the fixes to any consumers and before 25.10 release we will fully remove the old `Policy`.  We need an issue created to remind us of this
> * The UnitPolicy currently targets all the units of the juju application to which the UnitPolicy is added. This is correct, yes?
* correct
> * Adding a UnitPolicy kind of gives the source app a free ticket to access all endpoints of the target workload.
* only if they don't specify an `Endpoint`, right?
> * Also, if the target app does not have any AppPolicy, UnitPolicy also lets the source app to access the target service without any restriction (as mentioned by @ca-scribner in [Enable creating `AuthorizationPolicies` that bind to a workload not a Service (as in an L4 policy) #76](https://github.com/canonical/istio-beacon-k8s-operator/issues/76) since we dont have that allow nothing policy for waypoints yet). Was that supposed to be part of this PR?
* allow-nothing policy should not be part of this PR.  We'll have to add that separately.  But we should test this with an allow-nothing policy in place to ensure the policies we're creating here work

lib/charms/istio_beacon_k8s/v0/service_mesh.py
```diff
@@ -83,8 +76,11 @@ def __init__(self, *args):
 ```
 
 This example creates two policies:
-- When related over the `metrics` relation allow the related application to `GET` this application's `/metrics` endpoint on the specified port
-- When related over the `data` relation allow the relation application to `GET` this application's `/data` endpoint on the specified port
+- An unit policy - When related over the `metrics` relation allow the related application to access the units (via the unit/workload ip or FQDN) of this application without any restriction.
```
ca-scribner, line 79: I think mentioning IP or FQDN might be confusing here.  What about just saying the app's service or workload?

```suggestion
- A unit policy - When related over the `metrics` relation, allow the related application to access this application's unit Pods directly without any restriction.
```

lib/charms/istio_beacon_k8s/v0/service_mesh.py
```diff
@@ -83,8 +76,11 @@ def __init__(self, *args):
...
+- An app policy - When related over the `data` relation allow the relation application to `GET` this application's `/data` endpoint on the specified port (via the app/service ip or FQDN)
+
+An UnitPolicy creates a less restrictive policy that allows complete access but only to the units of the charm via individual unit addresses.
```
ca-scribner, line 82:
```suggestion
A UnitPolicy creates a less restrictive policy that allows complete access but only to the units of the charm via individual unit addresses.
```

lib/charms/istio_beacon_k8s/v0/service_mesh.py
```diff
@@ -83,8 +76,11 @@ def __init__(self, *args):
 ```
 
 This example creates two policies:
-- When related over the `metrics` relation allow the related application to `GET` this application's `/metrics` endpoint on the specified port
-- When related over the `data` relation allow the relation application to `GET` this application's `/data` endpoint on the specified port
+- An unit policy - When related over the `metrics` relation allow the related application to access the units (via the unit/workload ip or FQDN) of this application without any restriction.
+- An app policy - When related over the `data` relation allow the relation application to `GET` this application's `/data` endpoint on the specified port (via the app/service ip or FQDN)
```
ca-scribner, line 80: I think mentioning IP or FQDN might be confusing here.  What about just saying the app's service or workload?

```suggestion
- An app policy - When related over the `data` relation, allow the related application to `GET` this application's `/data` endpoint on the specified port through the app's Kubernetes service
```

lib/charms/istio_beacon_k8s/v0/service_mesh.py
ca-scribner, line 79: nit: can we put the app policy first and the unit policy second (in these explanations and in the code example above).  In nearly all cases the app is the right thing for people to use, so let's have them discover that first

lib/charms/istio_beacon_k8s/v0/service_mesh.py
ca-scribner, line 82: There was a miscommunication here I think.  The `UnitPolicy` should also accept the optional `Endpoints` argument, and restrict on endpoints exactly the same as `AppPolicy` does.  The only difference between `AppPolicy` and `UnitPolicy` is that one binds to services via TargetRef and the other to Pods via Selector.  The `UnitPolicy` will not be any less restrictive

lib/charms/istio_beacon_k8s/v0/service_mesh.py
```diff
@@ -193,13 +191,44 @@ class Endpoint(pydantic.BaseModel):
...
+
+    relation: str
+    endpoints: List[Endpoint]
+    service: Optional[str] = None
+
+
+class UnitPolicy(pydantic.BaseModel):
+    """Data type for defining a policy for your charm unit."""
+
+    relation: str
```
ca-scribner, line 230: See above discussion - we should have `endpoints` here too

src/charm.py
```diff
@@ -319,54 +320,87 @@ def _build_authorization_policies(self, mesh_info: List[MeshPolicy]):
...
-                        Rule(
-                            from_=[  # type: ignore # this is accessible via an alias
-                                From(
-                                    source=Source(
-                                        principals=[
-                                            _get_peer_identity_for_juju_application(
-                                                policy.source_app_name, policy.source_namespace
-                                            )
-                                        ]
+            # L4 policy created for target Juju units (workloads)
```
ca-scribner, line 323: The validation code above here

```python
            target_service = policy.target_service or policy.target_app_name
            if policy.target_service is None:
                logger.info(
                    f"Got policy for application '{policy.target_app_name}' that has no target_service. "
                    f"Defaulting to application name '{target_service}'."
                )
```
only applies to App policies, since they're the only ones that have target_service.  Let's move them down into the `if`

ca-scribner: Actually, now that we've clarified that `UnitPolicy` will also support `Endpoints`, let's refactor to reduce the duplication.  All that's different between App and Unit policies is the targetRef vs selector, so lets do something like:

```python
if app_policy:
  targetRefs = ...
  selector = None
elif unit_policy:
  targetRefs = None
  selector = ...
else:
  raise
```

Then the rest of the code is common.  That way it's clear how almost everything is common between them

tests/integration/test_charm.py
```diff
@@ -149,24 +158,26 @@ async def test_service_mesh_relation(ops_test: OpsTest, model_on_mesh):
     # configure auto-join for the apps based on model_on_mesh
     await ops_test.model.applications["receiver1"].set_config({"auto-join-mesh": str(not model_on_mesh).lower()})
     await ops_test.model.applications["sender1"].set_config({"auto-join-mesh": str(not model_on_mesh).lower()})
-    # Do not set auto-join for sender2, as it is not part of the mesh anyway
+    await ops_test.model.applications["sender2"].set_config({"auto-join-mesh": str(not model_on_mesh).lower()})
+    # Do not set auto-join for sender3, as it is not part of the mesh anyway
     # await ops_test.model.applications["sender2"].set_config({"auto-join-mesh": str(not model_on_mesh).lower()})
```
ca-scribner, line 163: This `# await ...`  comment is now for `sender3`, although really we can just remove the comment

tests/unit/test_mesh_consumer.py
```diff
@@ -238,6 +280,25 @@ def test_relation_data_policies(policies, expected_data):
     )
 
 
+def test_policy_deprecation_warning():
```
ca-scribner, line 283: This seems like an oddly precise, short-lived test :D  But I'm not against it

ca-scribner (commented):
a few comments/changes to make, but overall it looks great!

src/charm.py
adhityaravi: I have left it as it is for now as I don't really see a big improvement in code readability of maintainability by moving the if..else block out of and before the policy creation code block. 

But if you strongly feel against it, I can change it.

lib/charms/istio_beacon_k8s/v0/service_mesh.py
```diff
@@ -78,13 +68,20 @@ def __init__(self, *args):
...
 ```
 
 This example creates two policies:
-- When related over the `metrics` relation allow the related application to `GET` this application's `/metrics` endpoint on the specified port
-- When related over the `data` relation allow the relation application to `GET` this application's `/data` endpoint on the specified port
+- An app policy - When related over the `data` relation, allow the related application to `GET` this application's `/data` endpoint on the specified port through the app's Kubernetes service.
+- A unit policy - When related over the `metrics` relation, allow the related application to access this application's unit Pods directly on the specified port without any other restriction. UnitPolicy does not support fine-grained access control on the methods and paths via `Endpoints`.
+
+An AppPolicy created a more fine-grained policy that can be used to control how the source application can communicate with the target application via the app address.
+A UnitPolicy creates a less restrictive policy that allows complete access but only to the units of the charm via individual unit addresses.
```
ca-scribner, line 84: I'm a soft vote for removing the "fine-grained" and "less restrictive" wording here.  

Yes, its true, but that's not a factor that's important when choosing between them here.  Users will need to choose Unit vs App based on whether the thing talking to them is talking directly to units or via their service to the app.  I think that should be the focus of the text here - clarifying which they should choose and why

src/charm.py
```diff
@@ -312,61 +313,118 @@ def _build_authorization_policies(self, mesh_info: List[MeshPolicy]):
...
+            if policy.target_type == PolicyTargetType.unit:
+                # if the mesh policy of type unit contain any of the L7 attributes, warn and dont create the policy
+                valid_unit_policy = not any(
+                    endpoint.methods or endpoint.paths or endpoint.hosts
+                    for endpoint in policy.endpoints
                 )
+                if not valid_unit_policy:
+                    logger.warning(
+                        f"UnitPolicy requested between {policy.source_app_name} and {policy.target_app_name} is not created as it contains some unallowed policy attributes."
+                        "Unallowed policy attributes for UnitPolicy include paths, methods and hosts"
```
ca-scribner, line 326:
```suggestion
                        f"UnitPolicy requested between {policy.source_app_name} and {policy.target_app_name} is not created as it contains disallowed policy attributes:"
                        " UnitPolicy cannot contain paths, methods, or hosts"
```

src/charm.py
```diff
@@ -312,61 +313,118 @@ def _build_authorization_policies(self, mesh_info: List[MeshPolicy]):
...
-            )
+                                    for endpoint in policy.endpoints
+                                ],
+                            ),
+                        ],
+                    ).model_dump(by_alias=True, exclude_unset=True, exclude_none=True),
+                )
+
+            # L7 policy created for target Juju applications (services)
+            else:
```
ca-scribner, line 372: This suggests that all other types are accepted, but really I think we only want .app here and to reject anything else.  wdyt?

```suggestion
            elif(policy.target_type == PolicyTargetType.app):
              ...
            else:
              raise
```

src/charm.py
```diff
@@ -392,7 +450,7 @@ def _construct_waypoint(self):
             metadata=Metadata(
                 name=self._waypoint_name,
                 namespace=self.model.name,
-                labels={"istio.io/waypoint-for": "all", **self._telemetry_labels},
+                labels={"istio.io/waypoint-for": "service", **self._telemetry_labels},
```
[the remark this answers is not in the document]
adhityaravi: Decided to merge this nonetheless and fix previous implementations as they break.

PR #105, opened 2025-08-19
fix: fix service mesh lib consumer databag permission issue

adhityaravi:
## Issue
Fixes issue #104 

## Solution
Issue was all the units were attempting to write to the local application databag. Fixed by putting the operation behind a leader guard.

## Testing Instructions

1. Deploy the istio stack with the fix
2. Deploy alertmanager-k8s
3. Scale alertmanager-k8s to 2

### DIFF
--- a/lib/charms/istio_beacon_k8s/v0/service_mesh.py
+++ b/lib/charms/istio_beacon_k8s/v0/service_mesh.py
@@ -158,7 +158,7 @@ from ops import CharmBase, Object, RelationMapping
 
 LIBID = "3f40cb7e3569454a92ac2541c5ca0a0c"  # Never change this
 LIBAPI = 0
-LIBPATCH = 7
+LIBPATCH = 8
 
 PYDEPS = ["lightkube", "pydantic"]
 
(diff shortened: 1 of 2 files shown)


lib/charms/istio_beacon_k8s/v0/service_mesh.py
```diff
@@ -362,6 +366,8 @@ def labels(self) -> dict:
         return json.loads(self._relation.data[self._relation.app]["labels"])
 
     def _on_mesh_broken(self, _event):
+        if not self._charm.unit.is_leader():
```
adhityaravi, line 369: There is no need to put this event handler behind a leader guard. But also there is no reason why all units should be allowed to do this. Any opinions on whether it should behind a leader guard or not and possible reasons?

adhityaravi (commented):
A quick question

lib/charms/istio_beacon_k8s/v0/service_mesh.py
adhityaravi: which it is atm, so im merging then

PR #109, opened 2025-08-27
feat: add an option to auto allow comm between units of the charm

adhityaravi:
## Issue
fixes issue #106 

## Solution

- The ServiceMeshConsumer now includes a bollean argument `auto_allow_intra_app_access`
- When set to True, the `build_mesh_policies` method automatically appends an `UnitPolicy` with from the target apps service account selecting all the pods with the target app  label

## Testing Instructions
Can be tested using scaled Service Mesh Tester with more than 1 unit by setting the `auto_allow_intra_app_access` in the ServiceMeshConsumer to true and curling another units pod url from a unit.

### DIFF
--- a/lib/charms/istio_beacon_k8s/v0/service_mesh.py
+++ b/lib/charms/istio_beacon_k8s/v0/service_mesh.py
@@ -158,7 +158,7 @@ from ops import CharmBase, Object, RelationMapping
 
 LIBID = "3f40cb7e3569454a92ac2541c5ca0a0c"  # Never change this
 LIBAPI = 0
-LIBPATCH = 8
+LIBPATCH = 9
 
 PYDEPS = ["lightkube", "pydantic"]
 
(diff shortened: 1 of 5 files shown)


adhityaravi:
A few questions - 

- The `auto_allow_intra_app_access` is not dynamic meaning for ex. it cannot be parametrized as a charm config to allow the user to enable ti and disable it neither does it observe any juju event. IMO it wasnt necessary. 
- The `auto_allow_intra_app_access` name is debatable - I dont like it the best but I couldnt think of a better alternative that wasnt too long.  

Opinions on both?

lib/charms/istio_beacon_k8s/v0/service_mesh.py
```diff
@@ -265,6 +265,7 @@ def __init__(
         cross_model_mesh_provides_name: str = "provide-cmr-mesh",
         policies: Optional[List[Union[Policy, AppPolicy, UnitPolicy]]] = None,
         auto_join: bool = True,
+        restrict_cross_unit_communication: bool = False,
```
dstathis, line 268: Let's make this name a bit shorter. `peer_communication` is my suggestion.

dstathis: Also I think it should be an allow option and default to false. In other words, apps should not be able to talk to themselves unless they set this to True.

dstathis (approved):

lib/charms/istio_beacon_k8s/v0/service_mesh.py
adhityaravi: I like `peer_communication` the best of all the suggestion. I will go with that.

PR #112, opened 2025-09-09
test:  peer communication

adhityaravi:
## Issue
fixes issue #106 

## Solution
No additional pattern is implemented, the charm dev will be encouraged to add a `UnitPolicy`  for the required ports on the peer relation endpoint. 

So this PR simply adds a test to check such a peer policy works as intended


tests/integration/test_charm.py
```diff
@@ -264,6 +267,51 @@ async def test_service_mesh_relation(ops_test: OpsTest, model_on_mesh):
...
+    )
+
+    # sender2 with default config should allow cross unit communication
+    assert_request_returns_http_code(
+        ops_test.model.name,
+        "sender2/0",
+        f"http://sender2-1.sender2-endpoints.{ops_test.model.name}.svc.cluster.local:8080/foo",
+        code=200,
+    )
+
```
ca-scribner, line 314: everything here works great, but imo we can simplify the test logic a little 

deploying senders is cheap, so what about deploying an extra sender (`sender-scaled`?) and scaling that one.  Then things are just decoupled a tiny bit more

Also, rather than deploying it with `peer=false`, what about this:
* deploy sender-scaled with peer=true, scale=1
* scale to 2
* test peer to peer communication (should work)
* set config peer=false
* test peer to peer communication (should fail)

I think that'll make the test a bit easier to follow rather than reusing two of the old senders.  wdyt?

PR #118, opened 2025-10-01
feat: allow custom selectors in MeshPolicy

### LINKED ISSUE #117: Allow PolicyResourceManager to create policies with custom selector labels
### Enhancement Proposal

Currently the PolicyResourceManager can only create policies between a specific source and a target. But most practical use cases might need creating a policy between a source and more than one matching target (coordinated-worker) which can be addressed by letting the MeshPolicy object define the workloadSelectors instead of hardcoding the selector labels.

adhityaravi:
## Issue
fixes #117 

## Solution
`MeshPolicy` is extended to define `target_workload_selector` with some additonal validations to make sure that the policy obj can only be instantiated with a sane policy


lib/charms/istio_beacon_k8s/v0/service_mesh.py
```diff
@@ -281,12 +281,48 @@ class MeshPolicy(pydantic.BaseModel):
 
     source_app_name: str
     source_namespace: str
-    target_app_name: str
```
ca-scribner, line 284: This is awkward.  This was previously a positional arg, so moving it down the list to kwarg is a breaking API change.  Anyone using this will be broken

Do you know if anything outside this charm instantiates these objects?  I did a quick github search and only see instantiations in this lib or in beacon's tests, so I think its ok to break it... wdyt?

lib/charms/istio_beacon_k8s/v0/service_mesh.py
```diff
@@ -281,12 +281,48 @@ class MeshPolicy(pydantic.BaseModel):
...
+            self._validate_unit_policy()
+        return self
+
+    def _validate_app_policy(self) -> None:
+        """Validate app-targeted policy constraints."""
+        if not any([self.target_app_name, self.target_service]):
+            raise ValueError(
+                f"Bad policy configuration. Neither target_app_name nor target_service "
+                f"specified for MeshPolicy with target_type {self.target_type}"
+            )
```
ca-scribner, line 306: do we allow `all([app_name, service])`?  I remember us talking a bit before about this, but can't remember what we said.  I guess in terms of the policy object itself, defining both is valid, but in terms of istio *using* a policy that would be invalid (or at least redundant and we warn users)?

adhityaravi: Out of context for this PR to remove this. But it seems redundant to have this `target.service` and should be removed. I will for now add a warning regarding the redundancy.

lib/charms/istio_beacon_k8s/v0/service_mesh.py
adhityaravi: done

adhityaravi: and added a default for `endpoints`

lib/charms/istio_beacon_k8s/v0/service_mesh.py
adhityaravi: done

lib/charms/istio_beacon_k8s/v0/service_mesh.py
```diff
@@ -184,7 +185,7 @@ def __init__(self, *args):
 
 LIBID = "3f40cb7e3569454a92ac2541c5ca0a0c"  # Never change this
 LIBAPI = 0
-LIBPATCH = 10
+LIBPATCH = 11
```
ca-scribner, line 188: Because I was a jerk and merged #119 before you merged this :)

```suggestion
LIBPATCH = 12
```

adhityaravi: done!

PR #154, opened 2026-02-05
fix(lib): makes the label patcher least intrusive

### LINKED ISSUE #133: The `reconcile_charm_labels` doesnt handle k8s api conflict errors
### Bug Description

The `reconcile_charm_labels`  tries to patch the resources with the required labels. THe problem is as soon as the api is unable to patch the resource for reasons like but not ltd to

1. the k8s resource was patched by someone else and still settling down
2. lol, cant thing of any other reason

it raises an error causing the charm unit to go into a momentary error state until the next reconciliation. Rather, this should be handled in the label patching itself 

1. either by using the strategic merge patch type and/or
2. trying to patch the resource with an exponential backoff

### To Reproduce

bla

### Environment

bla

### Relevant log output

```shell
bla
```

### Additional context

_No response_

adhityaravi:
## Issue
Fixes #133 

## Context
The patching fn that patches labels on the consumer charms was basically fetching the entire resource description, and patching the labels and reapplying the whole resource description. This is an intrusive approach and causes issues when someone else patches a different part of the resource between us getting the resource definition and patching the resource with our labels. (happens with the coordinated workers when the `cosl` is applying the resource limits and we're applying the labels for ex.)

## Solution
This PR refactors this intrusive patch operation to only patch the labels we own. We don't really care about other resource definitions. This also prevents 409s with the k8s API because we own the label. This is also a reason we should not retry the patching operation because that might mask an actual issue of someone else editing service mesh labels.

## Testing Instructions
Just a refactor. just deploy anything with mesh and make sure the consumer charms are on the mesh with the required labels.

### DIFF
--- a/lib/charms/istio_beacon_k8s/v0/service_mesh.py
+++ b/lib/charms/istio_beacon_k8s/v0/service_mesh.py
@@ -180,7 +180,7 @@ POLICY_RESOURCE_TYPES = {
 
 LIBID = "3f40cb7e3569454a92ac2541c5ca0a0c"  # Never change this
 LIBAPI = 0
-LIBPATCH = 18
+LIBPATCH = 19
 
 PYDEPS = [
     "lightkube",
@@ -653,10 +653,7 @@ def reconcile_charm_labels(client: Client, app_name: str, namespace: str,  label
         labels: A dictionary of labels to set on the Charm's Kubernetes objects. Any labels that were previously created
                 by this method but omitted in `labels` now will be removed from the Kubernetes objects.
     """
-    patch_labels = {}
-    patch_labels.update(labels)
-    stateful_set = client.get(res=StatefulSet, name=app_name)
-    service = client.get(res=Service, name=app_name)
+    patch_labels: Dict[str, Optional[str]] = dict(labels)
     try:
         config_map = client.get(ConfigMap, label_configmap_name)
     except httpx.HTTPStatusError as e:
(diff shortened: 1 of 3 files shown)


lib/charms/istio_beacon_k8s/v0/service_mesh.py
```diff
@@ -670,19 +667,22 @@ def reconcile_charm_labels(client: Client, app_name: str, namespace: str,  label
...
+
+    # Do a least intrusive patch.
+    # This minimal approach eliminates the chance of 409 conflicts with other actors modifying the resources.
+    # Retrying here is a bad idea as we WANT to get a 409 when someone else patches OUR labels. We shouldn't mask that.
+    client.patch(res=StatefulSet, name=app_name, obj={
+        "spec": {"template": {"metadata": {"labels": patch_labels}}}
+    })
+    client.patch(res=Service, name=app_name, obj={
+        "metadata": {"labels": patch_labels}
+    })
```
dstathis: Why does this make eliminate the possibility of a 409? And if it does how does the next sentence make sense. Could you add a bit more explanation in the comment?

dstathis: Ah I see now. Maybe replace "Do a least intrusive patch" with "Patch only the labels rather than the whole object".

dstathis (approved):

lib/charms/istio_beacon_k8s/v0/service_mesh.py
adhityaravi: gotcha

adhityaravi: done

PR #156, opened 2026-02-18
feat: adds an enabled property to the mesh lib

### LINKED ISSUE #126: the ServiceMeshConsumer should provide a `is_active` method
### Enhancement Proposal

Currently the only way to know if a service mesh is connected to a charm or not is using the `mesh_type` property. which is not not obvious. To overcome this, we need to add a `is_active()` method or similar that tells if the service mesh connection exists more explicilty

adhityaravi:
## Issue
Fixes #126 

## Solution
instead of `is_active`, a property called `enabled` was added to indicate mesh connection

### DIFF
--- a/lib/charms/istio_beacon_k8s/v0/service_mesh.py
+++ b/lib/charms/istio_beacon_k8s/v0/service_mesh.py
@@ -456,6 +456,13 @@ class ServiceMeshConsumer(Object):
             return {}
         return app_data.labels
 
+    @property
+    def enabled(self) -> bool:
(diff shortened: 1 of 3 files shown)


dstathis:
I think we should be checking if the relation exists, not whether there is data in it. For example checking if self.relations.<relation-name> exists.

adhityaravi:
@dstathis you are right. changed it so we are simply checking if the relation exists

dstathis (approved):

PR #159, opened 2026-03-30
test: migrate to pytest jubilant 2

james-garner-canonical:
## Issue

The `pytest-jubilant` [2.0 release](https://github.com/canonical/pytest-jubilant/releases/tag/v2.0.0) breaks this repo's tests.

## Solution

This PR migrates the tests to use the 2.0 API. This includes defining copies of the `pack` and `get_resources` helpers dropped from `pytest-jubilant`.

## Context

Charm Tech will maintain the `pytest-jubilant` 2.0 API going forward. We dropped `pack` and `get_resources` to scope the library more tightly to `pytest` + `jubilant`. In general, we think it's cleaner to avoid calling `charmcraft pack` inside your Python integration test suite.

## Testing Instructions

Integration tests should pass.

## Upgrade Notes

N/A.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,7 +29,7 @@ dev = [
     # Integration
     "juju",
     "jubilant",
-    "pytest-jubilant",
+    "pytest-jubilant>=2,<3",
     "httpx",
     "tenacity",
 ]
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -8,9 +8,8 @@ import shutil
 from pathlib import Path
 
 import pytest
-from helpers import istio_k8s
+from helpers import get_resources, istio_k8s, pack
 from jubilant import all_active
-from pytest_jubilant import get_resources, pack
 
 
 @pytest.fixture(scope="session")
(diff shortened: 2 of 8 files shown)


james-garner-canonical:
The [test failure](https://github.com/canonical/istio-beacon-k8s-operator/actions/runs/23727475415/job/69114150973?pr=159) highlights an interesting problem: `pytest-jubilant`'s temporary model names are on the long side now. In this case, `jubilant-1191dd0e-test-policy-resource-manager-istio-system`, where `jubilant-1191dd0e` is the temp model's prefix, `test-policy-resource-manager` comes from the module, and `istio-system` is the suffix requested in the integration tests. This long name leads to the following error:
```
lightkube.core.exceptions.ApiError: unable to parse requirement: values[0][app.kubernetes.io/instance]: Invalid value: "istio-k8s-jubilant-1191dd0e-test-policy-resource-manager-istio-system": must be no more than 63 bytes
```
Workarounds on your side would include using a shorter suffix (`istio` instead of `istio-system`?), or a shorter module name (`test_prm.py` instead of `test_policy_resource_manager.py`). However, I wonder if this indicates instead that the charm logic needs an update to handle long model names without error.

adhityaravi:
Thanks for the PR @james-garner-canonical! I have created an issue to look into this.

PR #163, opened 2026-04-09
Make 'provide-cmr-mesh' optional in charmcraft.yaml

rpbritton:
## Issue

`provide-cmr-mesh` is not marked as optional, indicating that it is required by default.

## Solution

Mark as optional.

## Context

`istio-beacon-k8s` can run in a non-CMR context, making this endpoint optional.

## Testing Instructions

N/A?

## Upgrade Notes

N/A


adhityaravi:
Thanks for the contribution @rpbritton!

@dstathis sounds reasonable to me and it follows the service-mesh interface pattern anyway. wdyt?

gycsaba96 (approved):

PR #172, opened 2026-05-19
chore: uses istio charmlibs from pypi

adhityaravi:
## Issue
Uses charmlibs from charmlibs pypi and utilities from `canonical-service-mesh` pypi package


src/charm.py
```diff
@@ -595,20 +606,5 @@ def format_labels(label_dict: Dict[str, str]) -> str:
...
-    """Generate telemetry labels for the application, ensuring it is always <=63 characters and usually unique."""
-    telemetry_key = charm_kubernetes_label(
-        model_name=model_name,
-        app_name=app_name,
-        prefix="charms.canonical.com/",
-        suffix=".telemetry",
-    )
-    return {
-        telemetry_key: "aggregated",
-    }
```
dstathis, line 609: Why are we getting rid of this? is it not used anywhere?

adhityaravi: It is a shared utility fn moved to the central utility package `canonical-service-mesh`

dstathis (approved):

PR #182, opened 2026-06-01
fix: include icon.svg in final the charm file

gycsaba96:
The icon file must be explicitly referenced to be included in the final .charm file.


adhityaravi (approved):
LGTM

PR #185, opened 2026-06-04
Add charm-related links

gycsaba96:
This is useful for a nice looking Charmhub page.

### DIFF
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -2,7 +2,16 @@ name: istio-beacon-k8s
 type: charm
 title: Istio Beacon
 summary: A beacon charm used for joining the Istio service mesh.
-description: A beacon charm used for joining the Istio service mesh.
+description: |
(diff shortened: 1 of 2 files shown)


charmcraft.yaml
```diff
@@ -2,7 +2,16 @@ name: istio-beacon-k8s
 type: charm
 title: Istio Beacon
 summary: A beacon charm used for joining the Istio service mesh.
-description: A beacon charm used for joining the Istio service mesh.
+description: |
+  The istio-beacon-k8s charm facilitated adding other charms to the Istio service mesh.
```
adhityaravi: Should it be facilitates instead of facilitated?

adhityaravi (commented):
LGTM

charmcraft.yaml
gycsaba96: Good catch, thank you!

adhityaravi (approved):

PR #186, opened 2026-06-04
chore: use the same tooling during PRs and releases

gycsaba96:


.github/workflows/release.yaml
```diff
@@ -22,3 +22,5 @@ jobs:
     with:
       default-track: dev
       automatically-retry-hooks: true # A fix for #8
+      provider: 'microk8s'
+      charmcraft-channel: 'latest/beta'
```
adhityaravi, line 27: Is there a reason we are using the `latest/beta` channel here? I think the risk should be stable.

gycsaba96: Yes, it defaults to [latest/candidate](https://github.com/canonical/observability/blob/3ab21836827cbc54bd7ea8e4e949300381ad658d/.github/workflows/charm-release.yaml#L18), but charmcraft 4.2.1 fails to build the tester charm because of [this issue](https://github.com/canonical/charmcraft/issues/2707). I will add a comment about it.

adhityaravi: Comments are usually forgotten, please open an issue in the repository so we can track it in jira.

gycsaba96: Done.

adhityaravi (approved):

